PR #359, opened 2025-12-03
fix(sdk): correct FAILURE_TOLERANCE_EXCEEDED behavior in concurrent operations

ParidelPooya:
Fix bug where tolerance checks occurred after ALL_COMPLETED check,
    causing incorrect completion reasons when failure thresholds were exceeded.
    
    Changes:
    - Fix logic ordering in getCompletionReason to check tolerance before completion
    - Add failure-threshold-exceeded examples for map and parallel operations
    - Create tests validating FAILURE_TOLERANCE_EXCEEDED behavior

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/src/examples/map/min-successful/map-min-successful.test.ts
+++ b/packages/aws-durable-execution-sdk-js-examples/src/examples/map/min-successful/map-min-successful.test.ts
@@ -32,10 +32,12 @@ createTests({
       expect(item0?.getStatus()).toBe(OperationStatus.SUCCEEDED);
       expect(item1?.getStatus()).toBe(OperationStatus.SUCCEEDED);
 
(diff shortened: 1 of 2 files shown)


packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/concurrent-execution-handler.ts
```diff
@@ -301,17 +337,50 @@ export class ConcurrencyController<Logger extends DurableLogger> {
         return false;
       };
 
-      const getCompletionReason = ():
+      const getCompletionReason = (
```
anthonyting: Is this the same function as above? Could we extract it to some common function?

ParidelPooya:
> Can we have some more assertions in the integ tests on what is returned from the result? Currently it just asserts the counts, but it's possible that the results aren't returned correctly.

Added more assertions to tests

anthonyting (approved):

PR #363, opened 2025-12-04
ci: remove version when syncing file

anthonyting:
*Issue #, if available:*

*Description of changes:*

Currently syncing the SDK is not working since incrementing the package version breaks the sync.

We can't sync with the version in the filename, since we would have to parse the version from the filename to be able to untar it. We can just sync without the version in the filename.

### DIFF
--- a/.github/workflows/sync-sdk.yml
+++ b/.github/workflows/sync-sdk.yml
@@ -1,7 +1,7 @@
 name: Sync SDK
 on:
   push:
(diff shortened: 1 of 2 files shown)


SilanHe (dismissed):
best to test this on ci/remove-sync-version to eliminate potential back and forth but it looks good to me in theory

SilanHe (approved):

PR #383, opened 2025-12-11
ci: fix concurrency groups and cleanup only when PR is closed or merged

anthonyting:
*Issue #, if available:*

*Description of changes:*

Sometimes the integration tests would fail if multiple commits were added at the same time. This is due to each commit cleaning up the functions, while the next commit needed them or deployed them again.

The fix is to set up better concurrency groups, and also to only cleanup functions/logs when the PR is closed/merged instead of as part of the integ test. This makes it possible to inspect the execution in the console while the PR is still active to help with debugging or other reasons.


bchampp:
We should look at leveraging https://github.com/aws-actions/aws-lambda-deploy for deploying these functions

PR #386, opened 2025-12-12
refactor(sdk): optimize finishedAncestors with finished runInChildContext tracking

ParidelPooya:
- Track finished runInChildContext in finishedAncestors
    - Simplified hasFinishedAncestor that only checks finishedAncestors
    - Enable map-min-successful and parallel-min-successful tests

### DIFF
--- a/packages/aws-durable-execution-sdk-js-testing/src/checkpoint-server/handlers/checkpoint-handlers.ts
+++ b/packages/aws-durable-execution-sdk-js-testing/src/checkpoint-server/handlers/checkpoint-handlers.ts
@@ -105,7 +105,7 @@ export function processCheckpointDurableExecution(
     createCheckpointToken(input.CheckpointToken),
   );
 
-  validateCheckpointUpdates(updates, storage.operationDataMap);
+  validateCheckpointUpdates(updates, storage.getAllOperationData());
 
   storage.registerUpdates(updates);
 


packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts
```diff
@@ -72,60 +70,37 @@ export class CheckpointManager implements Checkpoint {
   }
 
   /**
-   * Checks if a step ID or any of its ancestors has a pending completion
+   * Mark an ancestor as finished (for run-in-child-context operations)
```
anthonyting: Could we just use markOperationState instead? And just add to `this.finishedAncestors` when the state is marked as `OperationLifeCycle.COMPLETED`? Then we don't need to have an extra function to use

packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts
```diff
@@ -304,16 +271,6 @@ export class CheckpointManager implements Checkpoint {
...
       this.queue.shift();
-
-      if (this.hasFinishedAncestor(nextItem.stepId, nextItem.data)) {
-        log("⚠️", "Checkpoint skipped - ancestor finished:", {
-          stepId: nextItem.stepId,
-          parentId: nextItem.data.ParentId,
-        });
-        skippedCount++;
-        continue;
-      }
```
anthonyting: I think we should be checking hasFinishedAncestor inside processQueue, since queue processing is async, so it's possible that an ancestor finished before we got to this batch.

packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts
```diff
@@ -662,14 +612,7 @@ export class CheckpointManager implements Checkpoint {
...
-        if (this.hasPendingAncestorCompletion(op.stepId)) {
-          log(
-            "🧹",
-            `Cleaning up operation with completed ancestor: ${op.stepId}`,
-          );
-          this.cleanupOperation(op.stepId);
-          this.operations.delete(op.stepId);
-        }
+        // Note: Ancestor completion checking removed - operations will continue normally
       }
```
anthonyting: We can remove this logic

packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts
ParidelPooya: OperationLifeCycle needs a lot of extra meta data. That was my first iteration but complexity was much higher

anthonyting (dismissed):

Issue #387, opened 2025-12-12
[Bug]: Testing SDK fails with ES modules: `__dirname is not defined`


ZackKanter:
### Expected Behavior

Expect that `@aws/durable-execution-sdk-js-testing@1.0.0` would be compatible with ES modules.

### Actual Behavior

 Tests fail with:

```
  ReferenceError: __dirname is not defined
      at CheckpointWorkerManager.getWorkerPath
```

### Steps to Reproduce

https://github.com/ZackKanter/durable-functions-bug-repro

### SDK Version

1.0.0

### Node.js Version

22.x

### Is this a regression?

No

### Additional Context

I may be missing something, as it seems that the library is intended to be compatible with ESM: https://github.com/aws/aws-durable-execution-sdk-js/issues/180

willfarrell:
I was going to write a PR for this. But it looks like someone beat me to it. https://github.com/aws/aws-durable-execution-sdk-js/pull/390

PR #390, opened 2025-12-12
fix(testing-sdk): fix errors when running as esm

anthonyting:
*Issue #, if available:*

https://github.com/aws/aws-durable-execution-sdk-js/issues/387

*Description of changes:*

The testing SDK does not work in ESM due to a dependency on `__dirname` which is CJS-only. 

Solution:
- Adding ESM shims to the rollup config with `@rollup/plugin-esm-shim`. This adds `__dirname`, `__filename`, and `require` as global variables in our bundle (only `__dirname` is used currently).
- Fix the path to the worker file when running as ESM

Additional changes:
- Remove file system checking and use compile-time variables to determine worker path
- Add step and wait example running as ESM (in local mode using vitest, not yet running against the cloud)


lukehedger:
When can we expect a new release with this bug fix?

PR #396, opened 2025-12-17
test: Add getRemainingTimeInMillis support to testing library

### LINKED ISSUE #270: Support Context.getRemainingTimeInMillis in testing library
Currently it throws a not implemented error. We should at least partially implement this function instead of just throwing an error.

willfarrell:
Closes #270

update context.lambdaContext to return 900_000 (15min) instead of throwing an error.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/ADDING_EXAMPLES.md
+++ b/packages/aws-durable-execution-sdk-js-examples/ADDING_EXAMPLES.md
@@ -87,21 +87,20 @@ Create a test file in the same directory:
 
 ```typescript
 import { handler } from "./{example-name}";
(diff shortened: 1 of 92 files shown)


packages/aws-durable-execution-sdk-js-testing/src/test-runner/local/__tests__/invoke-handler.test.ts
anthonyting: Could you please also add a test in this file https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js-testing/src/test-runner/local/__tests__/integration/local-durable-test-runner.integration.test.ts with a simple example that calls `context.lambdaContext.getRemainingTimeInMillis()` and validates the result?

anthonyting (approved):

PR #401, opened 2025-12-18
ci: add durable integration test using capacity provider

anthonyting:
*Issue #, if available:*

*Description of changes:*

Setting up durable integration tests using Lambda managed instances.

This required a lot of changes to the deployment script, since deploying functions to a capacity provider requires much more time and polling for when the function is actually ready.

The actual test is just a basic durable step, and we can add more complex ones later.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/scripts/deploy-lambda.ts
+++ b/packages/aws-durable-execution-sdk-js-examples/scripts/deploy-lambda.ts
@@ -548,7 +548,7 @@ async function main(): Promise<void> {
                     reason: "Function is now active",
                   };
                 },
-                210,
+                900,
               );
 
               if (


packages/aws-durable-execution-sdk-js-examples/scripts/deploy-lambda.ts
```diff
@@ -17,6 +15,14 @@ import {
   ResourceConflictException,
   UpdateFunctionConfigurationCommandInput,
   DeleteFunctionCommand,
+  Architecture,
```
bchampp: I didn't review this too closely but I trust that the changes work based on the [passing workflow run]( https://github.com/aws/aws-durable-execution-sdk-js/actions/runs/21043483000/job/60511491254?pr=401#step:7:90).

I'm wondering if at this point we should just write a CDK/CloudFormation template and use a parameter (to pass in the PR#) instead of maintaining our own version of the deployment script? It seems like a lot of maintenance work especially if we're adding more feature compatibility test coverage.

bchampp (approved):

PR #412, opened 2026-01-09
chore(sdk): Removes unnecessary throw statements

lmammino:
*Issue #, if available:*

https://github.com/aws/aws-durable-execution-sdk-js/discussions/411

*Description of changes:*

Removes unnecessary exception handling patterns in `with-durable-execution.ts`:

## 1. Simplified `validateDurableExecutionEvent`

  The previous implementation threw an error inside a try block and immediately caught all exceptions to replace with a different message:

  ```typescript
  function validateDurableExecutionEvent(event: unknown): void {
    try {
      const eventObj = event as Record<string, unknown>;
      if (!eventObj?.DurableExecutionArn || !eventObj?.CheckpointToken) {
        throw new Error("Missing required durable execution fields");
      }
    } catch {
      const msg = `Unexpected payload provided to start the durable execution...`;
      throw new Error(msg);
    }
  }
```

This was problematic because:

- The specific validation error message was never visible to callers
- Type assertions don't throw at runtime and optional chaining prevents most errors
- The `try/catch` was effectively redundant

Now simplified to directly throw the user-facing error without the wrapper.

## 2. Removed redundant catch block in `withDurableExecution`

The function contained a catch block that simply re-threw the error:

```ts
  try {
    response = await runHandler(...);
    return response;
  } catch (err) {
    throw err;
  }
```

This serves no purpose: errors propagate naturally without it. Replaced with a direct return of `runHandler()`.

---

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
+++ b/packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
@@ -268,15 +268,12 @@ async function runHandler<
  * Validates that the event is a proper durable execution input
  */
 function validateDurableExecutionEvent(event: unknown): void {
-  try {
-    const eventObj = event as Record<string, unknown>;
-    if (!eventObj?.DurableExecutionArn || !eventObj?.CheckpointToken) {
-      throw new Error("Missing required durable execution fields");
-    }
-  } catch {
-    const msg = `Unexpected payload provided to start the durable execution. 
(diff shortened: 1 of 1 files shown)


lmammino:
I understand this was failing some commit lint rules. I have rebased and (hopefully) fixed the commit message. Let me know if this needs more work before it can be merged.

anthonyting (approved):

Issue #466, opened 2026-03-05
[Feature]: Support tenantId in context.invoke() for Lambda tenant isolation


blimmer:
## What would you like?

AWS Lambda's [tenant isolation mode](https://docs.aws.amazon.com/lambda/latest/dg/tenant-isolation.html) (launched November 2025) allows multi-tenant SaaS applications to achieve execution environment isolation per tenant. When invoking a tenant-isolated Lambda function, a `TenantId` must be provided with the invocation request - omitting it causes the invocation to fail.

Currently, `context.invoke()` does not expose a way to pass a `TenantId` to the downstream function. The `InvokeConfig` interface only supports serialization options (`payloadSerdes`, `resultSerdes`), and the `ChainedInvokeOptions` sent to the checkpoint service hardcodes only `FunctionName`.

This means durable functions cannot invoke tenant-isolated Lambda functions, which is a blocker for customers adopting both features together.

**Desired behavior:**

```typescript
const result = await context.invoke(
  "process-payment",
  "arn:aws:lambda:us-east-1:123456789012:function:payment-processor:1",
  { amount: 100, currency: "USD" },
  { tenantId: "tenant-abc-123" },
);
```

## Possible Implementation

The Lambda service model (`ChainedInvokeOptions` shape) already defines `TenantId` as an optional field — the backend supports it. The change is limited to the SDK's user-facing API:

1. Add `tenantId?: string` to `InvokeConfig` in `src/types/invoke.ts`.
2. Pass `config.tenantId` through to `ChainedInvokeOptions.TenantId` in `src/handlers/invoke-handler/invoke-handler.ts`.
3. Update the local testing package to recognize and pass through `TenantId` in invoke operations.
4. Add tests covering the new parameter.

## Is this a breaking change?

No

## Does this require an RFC?

No

## Additional Context

- [AWS Lambda Tenant Isolation docs](https://docs.aws.amazon.com/lambda/latest/dg/tenant-isolation.html)
- [Invoking Lambda functions with tenant isolation](https://docs.aws.amazon.com/lambda/latest/dg/tenant-isolation-invoke.html)
- [Lambda Invoke API — TenantId parameter](https://docs.aws.amazon.com/lambda/latest/api/API_Invoke.html)
- The `ChainedInvokeOptions` shape in `.github/model/lambda.json` already includes `TenantId` as an optional member alongside `FunctionName`.

PR #467, opened 2026-03-05
feat(sdk): support tenantId in context.invoke() for tenant isolation

blimmer:
*Issue #, if available:* Resovles #466

*Description of changes:*

Add `tenantId` option to `InvokeConfig` so durable functions can invoke tenant-isolated Lambda functions. The `TenantId` is passed through to `ChainedInvokeOptions`, which the backend already supports.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,6 +64,8 @@ jobs:
 
   integration-tests:
     needs: build
+    # Skip integration tests for PRs from forked repos (no access to secrets)
+    if: github.event.pull_request.head.repo.full_name == github.repository || github.event_name != 'pull_request'
     permissions:
       contents: read
       id-token: write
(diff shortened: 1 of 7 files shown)


packages/aws-durable-execution-sdk-js/src/types/invoke.ts
```diff
@@ -9,4 +9,6 @@ export interface InvokeConfig<I, O> {
   payloadSerdes?: Serdes<I>;
   /** Serialization/deserialization configuration for result data */
   resultSerdes?: Serdes<O>;
+  /** Tenant identifier for invoking tenant-isolated Lambda functions */
+  tenantId?: string;
```
blimmer: This felt like the simplest backward-compatible place to set this. If we were OK with a break change, it might make sense to have more of an "options bag" that takes all current and future `ChainedInvokeOptions` (omitting `FunctionName`).

Issue #492, opened 2026-03-27
[Bug]: Parallel invokes fail to suspend promptly after checkpoint queue drains


ayushiahjolia:
### Expected Behavior

When the fast invoke completes and the remaining invokes are all in IDLE_AWAITED state with an empty checkpoint queue, the function should suspend promptly (within ~20ms cooldown) and stop consuming Lambda compute time while waiting for the remaining invokes.

### Actual Behavior

The function does not suspend. Instead it falls through to forceRefreshAndCheckStatus polling (1s initial, backing off to 10s, up to 15 minutes), burning Lambda compute time unnecessarily until each remaining invoke completes. The termination check is never re-evaluated after the checkpoint queue drains because processQueue() calls notifyQueueCompletion() but not checkAndTerminate().

### Steps to Reproduce

1. Create a durable function that uses context.parallel() (or context.map()) with multiple context.invoke() branches
2. Ensure at least one invoke completes significantly faster than the others
3. Invoke the function asynchronously and observe execution behavior

### SDK Version

1.0.0

### Node.js Version

22.x

### Is this a regression?

No

PR #503, opened 2026-04-15
fix(sdk): re-evaluate termination after checkpoint queue drains

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-js/issues/492

*Description of changes:*

When `context.parallel()` dispatches multiple context.invoke() branches, the function correctly suspends after all invokes are dispatched. However, when the first invoke completes and triggers a resume, the runtime stays active polling for remaining branches instead of re-suspending.

Root cause - processQueue()'s finally block notifies queue completion but doesn't call checkAndTerminate(). When a completed branch checkpoints SUCCEED (queued via fire-and-forget), the pending branches' markOperationAwaited→checkAndTerminate()` calls see a non-empty queue and skip termination. After the queue drains, nobody re-evaluates.

Fix - this.checkAndTerminate()afterthis.notifyQueueCompletion()inprocessQueue`'s finally block.

Testing - unit tests and integ tests

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/src/examples/parallel/invoke/parallel-invoke.history.json
+++ b/packages/aws-durable-execution-sdk-js-examples/src/examples/parallel/invoke/parallel-invoke.history.json
@@ -1,0 +1,252 @@
+[
+  {
+    "EventType": "ExecutionStarted",
+    "EventId": 1,
+    "Id": "23588ea1-7c59-49f1-88f6-5583f4cafc12",
+    "EventTimestamp": "2026-04-14T22:24:08.163Z",
+    "ExecutionStartedDetails": {
+      "Input": {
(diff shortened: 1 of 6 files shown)


ParidelPooya (approved):

anatolzak:
Thank you for this fix!

We got bitten by exactly this, a `ctx.map` of `waitForCallbacks` where, once the first branch came back, the function just kept polling for the rest instead of going back to sleep. So every callback wait ran the invocation all the way out to its timeout. On a busy workflow that got expensive fast. about $3k before we figured out what was going on, and on track for ~$7k/month if we hadn't caught it.

Took us a while to chase down, so stumbling onto this PR was a huge relief, it was exactly what we were seeing. Thanks again for getting it in!

PR #505, opened 2026-04-21
feat(sdk): add retryOperation helper for durable operations

ParidelPooya:
**What**

Adds a new withRetry helper to @aws/durable-execution-sdk-js that runs a chunk of durable-execution logic with automatic retries and configurable backoff. Semantically it's runInChildContext with a retry policy
   wrapped around it — on failure the whole function body is re-run from the beginning until it succeeds, retries are exhausted, or the strategy decides to stop.

**Why**
context.step() has first-class retry support via StepConfig.retryStrategy, but other durable primitives (waitForCallback, invoke, waitForCondition, etc.) don't — and they can't be nested inside a step. Today
   every team hand-rolls the same pattern:
```
   let attempt = 0;
   while (attempt < 3) {
     attempt++;
     try {
       return await context.waitForCallback(`approval-${attempt}`, submitter, {
         timeout: { hours: 24 },
       });
     } catch (err) {
       if (attempt >= 3) throw err;
       await context.wait(`backoff-${attempt}`, { seconds: 2 ** attempt });
     }
   }
```
   It's easy to get subtly wrong (forgetting context.wait for backoff, using setTimeout, accidentally nesting durable ops inside a step). withRetry turns it into a one-liner and reuses the exact same retryStrategy
    shape as StepConfig — so createRetryStrategy and retryPresets work out of the box.

**Usage**
```
   import { withRetry, createRetryStrategy } from "@aws/durable-execution-sdk-js";

   const decision = await withRetry(
     context,
     "approval",
     (ctx, attempt) =>
       ctx.waitForCallback<"APPROVED" | "REJECTED">(
         `approval-${attempt}`,
         (callbackId) => sendApprovalEmail(event.email, callbackId),
         { timeout: { hours: 24 } },
       ),
     {
       retryStrategy: createRetryStrategy({
         maxAttempts: 3,
         initialDelay: { seconds: 2 },
         backoffRate: 2,
       }),
     },
   );
```
   Anonymous overload (no name, no child-context grouping name) is also supported. The body can contain multiple durable operations — it's not limited to a single op.

**Behavior**

   - Runs func(ctx, attempt) starting at attempt = 1.
   - On rejection → calls retryStrategy. If shouldRetry: true, context.waits for the returned delay, then re-runs the whole function.
   - On fulfillment → returns the value.
   - On the final failure → rethrows (see "Error shape" below).
   - By default wraps the loop in context.runInChildContext for isolation and so all attempts + backoff waits appear as a single grouped operation in execution history. Opt out via wrapWithRunInChildContext: false.
   - Optional `childContextConfig` is forwarded to the wrapping runInChildContext (e.g. serdes, subType, errorMapper, virtualContext).

**Public API**

Three new exports from the SDK barrel:
```
   export function withRetry<T>(context, name, func, config): Promise<T>;
   export function withRetry<T>(context, func, config): Promise<T>;

   export interface WithRetryConfig<T> {
     retryStrategy: (error: Error, attemptCount: number) => RetryDecision;
     wrapWithRunInChildContext?: boolean;         // default: true
     childContextConfig?: ChildConfig<T>;
   }

   export type RetryableFunc<T> = (
     context: DurableContext,
     attempt: number, // 1-based
   ) => Promise<T>;
```
**Error shape**

   Because the loop is wrapped in runInChildContext by default, the final failure is rethrown as a ChildContextError with the original error on .cause. If you're doing instanceof checks on the error, either inspect
   error.cause or set wrapWithRunInChildContext: false to get the raw error.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/template.yml
+++ b/packages/aws-durable-execution-sdk-js-examples/template.yml
@@ -1373,6 +1373,31 @@ Resources:
           DURABLE_EXAMPLES_VERBOSE: "true"
     Metadata:
       SkipBuild: "True"
+  ParallelInvoke:
+    Type: AWS::Serverless::Function
+    Properties:
+      FunctionName: ParallelInvoke-22x-NodeJS-Local
+      CodeUri: ./dist
+      Handler: parallel-invoke.handler
+      Runtime: nodejs22.x
+      Architectures:
+        - x86_64
+      MemorySize: 128
+      Timeout: 60
+      Role:
+        Fn::GetAtt:
+          - DurableFunctionRole
+          - Arn
+      DurableConfig:
+        ExecutionTimeout: 60
+        RetentionPeriodInDays: 7
+      Environment:
+        Variables:
+          AWS_ENDPOINT_URL_LAMBDA: http://host.docker.internal:5000
+          DURABLE_VERBOSE_MODE: "false"
+          DURABLE_EXAMPLES_VERBOSE: "true"
+    Metadata:
+      SkipBuild: "True"
   ParallelMinSuccessful:
     Type: AWS::Serverless::Function
     Properties:
@@ -2148,6 +2173,31 @@ Resources:
           DURABLE_EXAMPLES_VERBOSE: "true"
     Metadata:
       SkipBuild: "True"
+  WaitConfigurable:
+    Type: AWS::Serverless::Function
+    Properties:
+      FunctionName: ConfigurableWait-22x-NodeJS-Local
+      CodeUri: ./dist
+      Handler: wait-configurable.handler
+      Runtime: nodejs22.x
+      Architectures:
+        - x86_64
+      MemorySize: 128
+      Timeout: 60
+      Role:
+        Fn::GetAtt:
+          - DurableFunctionRole
+          - Arn
+      DurableConfig:
+        ExecutionTimeout: 60
+        RetentionPeriodInDays: 7
+      Environment:
+        Variables:
+          AWS_ENDPOINT_URL_LAMBDA: http://host.docker.internal:5000
+          DURABLE_VERBOSE_MODE: "false"
+          DURABLE_EXAMPLES_VERBOSE: "true"
+    Metadata:
+      SkipBuild: "True"
   WaitNamed:
     Type: AWS::Serverless::Function
     Properties:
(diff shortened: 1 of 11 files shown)


packages/aws-durable-execution-sdk-js-examples/template.yml
```diff
@@ -1373,6 +1373,31 @@ Resources:
           DURABLE_EXAMPLES_VERBOSE: "true"
     Metadata:
       SkipBuild: "True"
+  ParallelInvoke:
```
SilanHe, line 1376: nit: is this `ParallelInvoke` and `WaitConfigurable` intended to be included in this PR?

SilanHe (dismissed):

SilanHe:
I think the PR looks good on its own, so I approved. However, I do think that it's worth asking for some feedback from customers like @embano1 suggested. Especially, prior to replicating the work for Python and Java.

FYI: @zhongkechen

packages/aws-durable-execution-sdk-js-examples/template.yml
ParidelPooya: Those are new integration tests added in this PR

SilanHe: It doesn't look like you've just defined them, were they not running previously?

Seems we're amending the PR here: https://github.com/aws/aws-durable-execution-sdk-js/commit/b524eb939596ecc7eaa8be7d3eb6768f6cffbbc1

SilanHe (approved):

packages/aws-durable-execution-sdk-js/src/utils/with-retry/index.ts
```diff
@@ -0,0 +1,251 @@
...
+      try {
+        return await func(ctx, attempt);
+      } catch (err) {
+        const decision = retryStrategy(err as Error, attempt);
+        if (!decision.shouldRetry) throw err;
+        const delay = decision.delay ?? { seconds: 1 };
+        if (name) {
+          await ctx.wait(`${name}-backoff-${attempt}`, delay);
+        } else {
+          await ctx.wait(delay);
```
yaythomas, line 240: why not name it anyway? `backoff-${attempt}`

packages/aws-durable-execution-sdk-js/src/utils/with-retry/index.ts
```diff
@@ -0,0 +1,251 @@
...
+   *
+   * @example
+   * ```typescript
+   * retryStrategy: (error, attempt) => ({
+   *   shouldRetry: attempt < 3,
+   *   delay: { seconds: 2 ** attempt },
+   * })
+   * ```
+   */
+  retryStrategy: (error: Error, attemptCount: number) => RetryDecision;
```
yaythomas, line 35: would be nice to have example w `createRetryStrategy`, this is one of the big benefits of this approach :-)

packages/aws-durable-execution-sdk-js/src/utils/with-retry/index.ts
```diff
@@ -0,0 +1,251 @@
...
+   *   property (standard `runInChildContext` behavior).
+   * - When `false`, the original error from the final failed attempt is
+   *   rethrown unchanged.
+   *
+   * Code that uses `instanceof` to branch on error type should account for
+   * this, or always inspect `error.cause`.
+   *
+   * @defaultValue true
+   */
+  wrapWithRunInChildContext?: boolean;
```
yaythomas, line 56: `wrapInChildContext`?

you could also collapse this into the `chjildContextConfig`, i.e treat existence/non-existence of that field as the bool, but I guess you want to apply the default ChildConfig w/o making the consumer specify it?

packages/aws-durable-execution-sdk-js/src/utils/with-retry/index.ts
ParidelPooya: Because we want to gice developers more control, if they don't use a name, then why should we use a name?

packages/aws-durable-execution-sdk-js/src/utils/with-retry/index.ts
ParidelPooya: +1 to. what you said, use it by default without forcing developers to mention chjildContextConfig

packages/aws-durable-execution-sdk-js/src/utils/with-retry/index.ts
ParidelPooya: Both examples we added have retryStrategy

embano1:
Great job, @ParidelPooya ! Can't wait to try!

Question: how do we ensure the other SDKs and [documentation](https://github.com/aws/aws-durable-execution-docs) stay consistent with those changes? cc/ @yaythomas

yaythomas:
other SDKs + docs tracking issues linked above

SilanHe:
https://github.com/aws/aws-durable-execution-docs/discussions/140

PR #514, opened 2026-04-29
feat(sdk): add configureSerdes for default serdes configuration

ParidelPooya:
Add context.configureSerdes() to set default serialization/deserialization behavior once for all operations, instead of passing serdes per-operation.

- Add SerdesConfig interface with defaultSerdes and defaultCallbackDeserializer
- Add configureSerdes() to DurableContext interface and DurableContextImpl
- Thread getDefaultSerdes through step, invoke, runInChildContext, and waitForCondition handlers
- Thread getDefaultCallbackDeserializer through createCallback and waitForCallback handlers; callbacks default to passthrough independently of defaultSerdes (must opt-in explicitly)
- Propagate serdes config to child contexts created by runInChildContext
- Add createFileSystemSerdes(basePath) helper for S3 Files mount integration
- Add configure-serdes example and integration test with history snapshot
- Export SerdesConfig and createFileSystemSerdes from package index

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/handlers/callback-handler/callback.ts
+++ b/packages/aws-durable-execution-sdk-js/src/handlers/callback-handler/callback.ts
@@ -30,7 +30,7 @@ export const createCallback = (
   createStepId: () => string,
   checkAndUpdateReplayMode: () => void,
   parentId?: string,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+
   getDefaultCallbackDeserializer?: () => AnySerdesDeserializer,
 ) => {
   return <T>(
--- a/packages/aws-durable-execution-sdk-js/src/handlers/callback-handler/callback.ts
+++ b/packages/aws-durable-execution-sdk-js/src/handlers/callback-handler/callback.ts
@@ -9,7 +9,7 @@ import {
 import { OperationStatus, OperationType } from "@aws-sdk/client-lambda";
 import { log } from "../../utils/logger/logger";
 import { Checkpoint } from "../../utils/checkpoint/checkpoint-helper";
-import { Serdes } from "../../utils/serdes/serdes";
+import { Serdes, AnySerdesDeserializer } from "../../utils/serdes/serdes";
 import { safeDeserialize } from "../../errors/serdes-errors/serdes-errors";
 import {
   CallbackError,
@@ -31,7 +31,7 @@ export const createCallback = (
   checkAndUpdateReplayMode: () => void,
   parentId?: string,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  getDefaultCallbackDeserializer?: () => Pick<Serdes<any>, "deserialize">,
+  getDefaultCallbackDeserializer?: () => AnySerdesDeserializer,
 ) => {
   return <T>(
     nameOrConfig?: string | undefined | CreateCallbackConfig<T>,
--- a/packages/aws-durable-execution-sdk-js/src/handlers/callback-handler/callback.ts
+++ b/packages/aws-durable-execution-sdk-js/src/handlers/callback-handler/callback.ts
@@ -30,6 +30,8 @@ export const createCallback = (
   createStepId: () => string,
   checkAndUpdateReplayMode: () => void,
   parentId?: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  getDefaultCallbackDeserializer?: () => Pick<Serdes<any>, "deserialize">,
 ) => {
   return <T>(
     nameOrConfig?: string | undefined | CreateCallbackConfig<T>,
@@ -46,7 +48,11 @@ export const createCallback = (
     }
 
     const stepId = createStepId();
-    const serdes = config?.serdes || createPassThroughSerdes<T>();
+    const serdes =
+      config?.serdes ||
+      (getDefaultCallbackDeserializer
+        ? getDefaultCallbackDeserializer()
+        : createPassThroughSerdes<T>());
 
     // Phase 1: Setup and checkpoint
     let isCompleted = false;
--- a/packages/aws-durable-execution-sdk-js/src/types/durable-context.ts
+++ b/packages/aws-durable-execution-sdk-js/src/types/durable-context.ts
@@ -1,5 +1,6 @@
 import { Context } from "aws-lambda";
 import { LoggerConfig } from "./logger";
+import { SerdesConfig } from "../utils/serdes/serdes";
 import { StepFunc, StepConfig } from "./step";
 import { ChildFunc, ChildConfig } from "./child-context";
 import { InvokeConfig } from "./invoke";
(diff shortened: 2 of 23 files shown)


packages/aws-durable-execution-sdk-js/src/utils/serdes/serdes.ts
```diff
@@ -44,6 +44,29 @@ export interface Serdes<T> {
...
+ * Configuration for default serdes behavior on a DurableContext.
+ *
+ * @public
+ */
+export interface SerdesConfig {
+  /**
+   * Default serdes used for step, runInChildContext, invoke, and waitForCondition results.
+   * Falls back to the built-in JSON serdes if not provided.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
ayushiahjolia: Nit: We can define this at one place and use it everywhere?
```
// eslint-disable-next-line @typescript-eslint/no-explicit-any
type AnySerdes = Serdes<any>;
```
like
```
defaultSerdes?: AnySerdes;
```

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Controls when data is written to the filesystem.
+ *
+ * - `ALWAYS`: Every value is written to a file; the checkpoint stores only a file pointer.
+ *   Best for consistently large payloads or when you want predictable checkpoint sizes.
+ *
+ * - `OVERFLOW`: Data is written inline (as JSON) unless it exceeds the durable function
+ *   checkpoint size limit (~256KB), in which case it overflows to a file.
+ *   Best for mixed workloads where most payloads are small.
+ *
```
ayushiahjolia, line 18: Is it feasible to add a test for this?

ayushiahjolia (dismissed):

packages/aws-durable-execution-sdk-js/src/utils/serdes/serdes.ts
ParidelPooya: Applied

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
ParidelPooya: Added

ayushiahjolia (dismissed):

packages/aws-durable-execution-sdk-js/src/types/durable-context.ts
```diff
@@ -848,4 +849,27 @@ export interface DurableContext<TLogger extends DurableLogger = DurableLogger> {
...
+   * ```typescript
+   * // Only steps use filesystem serdes; callbacks keep the default passthrough
+   * context.configureSerdes({ defaultSerdes: createFileSystemSerdes("/mnt/s3") });
+   *
+   * // Opt-in: both steps and callbacks use filesystem serdes
+   * const fsSerdes = createFileSystemSerdes("/mnt/s3");
+   * context.configureSerdes({ defaultSerdes: fsSerdes, defaultCallbackDeserializer: fsSerdes });
+   * ```
+   */
+  configureSerdes(config: SerdesConfig): void;
```
yaythomas, line 874: `setDefaultSerdes`?

ParidelPooya: We already have configureLogger, I want to keep the name similar.

packages/aws-durable-execution-sdk-js/src/handlers/callback-handler/callback.ts
```diff
@@ -46,7 +48,11 @@ export const createCallback = (
     }
 
     const stepId = createStepId();
-    const serdes = config?.serdes || createPassThroughSerdes<T>();
+    const serdes =
+      config?.serdes ||
+      (getDefaultCallbackDeserializer
+        ? getDefaultCallbackDeserializer()
+        : createPassThroughSerdes<T>());
```
yaythomas, line 55: doesn't this end up double-deserializing?

createCallback deserializes with defaultCallbackDeserializer

And then waitForCallback runs safeDeserialize on that same value?

Previously the inner callback was always on passthrough for waitForCallback.

yaythomas (changes requested):

packages/aws-durable-execution-sdk-js/src/handlers/callback-handler/callback.ts
ParidelPooya: Good catch. You're right that there was a double-deserialization risk. Here's how it works now:

   Without configureSerdes (existing behavior, unchanged):

   - createCallback uses passthrough (raw string through)
   - runInChildContext uses defaultSerdes (JSON round-trip)
   - Phase 2 uses passthrough → user gets the raw string

   With configureSerdes({ defaultCallbackDeserializer }) (new behavior):

   - createCallback is forced to passthrough (we inject it explicitly) → raw string preserved
   - runInChildContext is also forced to passthrough → raw string preserved through storage round-trip
   - Phase 2 applies defaultCallbackDeserializer exactly once on the raw string

   The key insight is that both createCallback and runInChildContext need passthrough when a defaultCallbackDeserializer is configured — otherwise the JSON round-trip in runInChildContext would cause an
   inconsistency between the first invocation and replay (on replay, handleCompletedChildContext deserializes the stored result, so phase 2 would receive an already-parsed object instead of a string).

   The passthrough injection is conditional — only applied when getDefaultCallbackDeserializer is set — so existing users who don't use configureSerdes see zero behavior change.

   A regression test was added in configure-callback-deserializer.test.ts that would fail without this fix: it sends JSON.stringify({ value: 42 }) as the callback result and asserts the handler receives { value: 42
   } (parsed object), not a string or a double-parse error.

yaythomas: > The passthrough injection is conditional — only applied when getDefaultCallbackDeserializer is set — so existing users who don't use configureSerdes see zero behavior change

https://github.com/aws/aws-durable-execution-sdk-js/blob/9b63cb8eb08d8ea130a2f7c213f02d174b036c92/packages/aws-durable-execution-sdk-js/src/context/durable-context/durable-context.ts#L486

```
    return this.withDurableModeManagement(() => {
      const waitForCallbackHandler = createWaitForCallbackHandler(
        this._executionContext,
        this.getNextStepId.bind(this),
        this.runInChildContext.bind(this),
        () => this._defaultCallbackDeserializer,
      );
```

This is not conditional, though?  passed unconditionally, getDefaultCallbackDeserializer always truthy inside createWaitForCallbackHandler.

So `runInChildContext` is always called with `{ subType, errorMapper, serdes: createPassThroughSerdes() }`, and then the child's result is now serialized with passthrough (`"str"` becomes `"str"`).

Pre-PR it went through `defaultSerdes` (`"str"` becomes `"\"str\""`)

Which doesn't necessarily matter, but it is changing the storage format and could have execution time impact if SDK ugprades?

ParidelPooya: Fixed: added a _customCallbackDeserializerSet flag that is only set to true when the user explicitly calls configureSerdes({ defaultCallbackDeserializer: ... }). The getter is only passed to
  createWaitForCallbackHandler when that flag is true, and child context propagation also respects it. Users who never call configureSerdes see zero behavior change.

yaythomas (approved):

Issue #510, opened 2026-04-25
[Bug]: context.map branches with waitForCallback do not progress after all callbacks signal; execution wedges until an unrelated timer resumes it


t-d-d:
### Expected Behavior

When every branch in a `context.map` completes its `ctx.waitForCallback` via `SendDurableExecutionCallbackSuccess`, the durable service should schedule an orchestrator resume to observe `context.map` completion and advance to the next `ctx.step`.

### Actual Behavior

After the durable service accepts the **last** branch's `SendDurableExecutionCallbackSuccess` (visible as a short 23–371 ms orchestrator Lambda invocation that logs no app-level error), the orchestrator goes silent for ~11–12 minutes. No further Lambda invocations for that execution, no errors in CloudWatch, `list-durable-executions-by-function` shows `Status: RUNNING` with no `EndTimestamp`.

Only when a sibling `ctx.wait({ minutes: 15 })` in a parallel branch (not related to the map) fires, the orchestrator is resumed. On that resume it walks the history, observes `context.map` is complete, and executes the remaining `ctx.step`s in ~8 seconds. **Two back-to-back reproductions showed the exact same 11-min silence pattern.**

Interleaved CloudWatch timeline for one reproduction (two concurrent executions A1 and A2, each a 2-branch map):

```
17:10:53  A1 LAMBDA-START req=28e1dbd5  (initial invoke; map dispatched)
17:11:10  A2 LAMBDA-START req=7c945785  (initial invoke; map dispatched)
17:11:34  CALLBACK-SIGNAL A1 branch=1           (SendDurableExecutionCallbackSuccess)
17:11:34  A1 LAMBDA-START req=43a63745  dur=23ms     ← service resumes, accepts signal
17:13:58  CALLBACK-SIGNAL A2 branch=1
17:13:58  A2 LAMBDA-START req=33337690  dur=194ms
17:14:00  CALLBACK-SIGNAL A1 branch=0  ** last branch **
17:14:00  A1 LAMBDA-START req=f30800df  dur=67ms     ← accepts signal, returns
17:14:52  CALLBACK-SIGNAL A2 branch=0  ** last branch **
17:14:52  A2 LAMBDA-START req=ddc9e57b  dur=371ms    ← accepts signal, returns

────── 11-MINUTE SILENCE — NO INVOCATIONS FOR EITHER EXECUTION ──────

[...]

PR #520, opened 2026-04-30
fix(testing-sdk): don't gate PENDING-recovery on hasScheduledFunction

ParidelPooya:
In `TestExecutionOrchestrator`, when the handler returns PENDING, a recovery path re-invokes the handler if the server has dirty operations that the handler may not have observed before it returned:

    if (!this.scheduler.hasScheduledFunction() && hasDirtyOperations) {
      this.scheduler.scheduleFunction(() => this.invokeHandler(...), ...);
    }

The `!hasScheduledFunction()` gate is wrong. It was likely intended to avoid double-scheduling, but it conflates "something is scheduled" with "that something will handle the pending updates" — which isn't true when the scheduled timer is unrelated, for example the timer for a sibling `ctx.wait` in a parallel branch.

Concrete failure shape (mirrors issue #510):

    parallel(
      { name: 'body',     func: ctx => map(..., waitForCallback) },
      { name: 'watchdog', func: ctx => ctx.wait(N) },
    )

Timeline without this fix:

 1. Initial invocation registers the map, the callbacks, and the watchdog's `ctx.wait`. `handleWaitUpdate` schedules a `setTimeout(..., N seconds)` via `TimerScheduler` for the watchdog.
 2. Termination fires with `WAIT_SCHEDULED` (correct — nothing has completed yet). Handler returns PENDING.
 3. Test calls `sendCallbackSuccess`. Callbacks become dirty on the server. `handleCallbackUpdate` schedules a 0ms `setTimeout` for the follow-up invocation.
 4. That 0ms timer fires while the first invocation is still mid-teardown, hits the `hasActiveInvocation()` guard, and is silently dropped.
 5. The first invocation completes. `hasDirtyOperations` is true — the callback updates are sitting in the server's dirty set. But `hasScheduledFunction()` is also true, because the watchdog's N-second timer is still in `runningTimers`. The gate blocks recovery and nothing else re-invokes.
 6. Execution sits idle until the watchdog's scheduled time arrives, then re-invokes, finishes the map in seconds, and — if `parallel` was configured with `minSuccessful: 1` — takes the watchdog's result while the body branch's work is abandoned.

Fix: re-invoke whenever `hasDirtyOperations` is true at PENDING time, without gating on `hasScheduledFunction`. Pre-existing scheduled functions (like the watchdog timer) are independent of checkpoint-sync state and will still fire on their own schedule.

Reproducers / regression tests added under
`packages/.../examples/map/`:

 - `wait-for-callback/` — bare `map + waitForCallback` (2 branches).
 - `wait-for-callback-many/` — 4 branches with staggered callbacks.
 - `wait-for-callback-parallel/` — the exact issue-#510 shape; asserts `elapsedMs < 1500` against a 4-second watchdog.
 - `wait-for-callback-parallel-nowatchdog/` — no-watchdog control.

Before this change the `wait-for-callback-parallel` test takes ~watchdog_delay ms (4000+ ms) and `parallel` resolves with only the watchdog's result. After the change it completes in ~300 ms with the body's result, as intended.

Scope of this fix
-----------------

This is a fix in the **testing library's local test orchestrator**, not in the core SDK. Investigation did not find a corresponding bug in the core SDK: when callbacks are delivered to the handler, the SDK's callback handler, checkpoint manager, and termination logic behave correctly.

The production issue reported in issue #510 (11–12 minute silence on real Lambda executions) likely has a structurally similar root cause in the Durable Execution service (which this repo does not include). That would need to be addressed separately by the service team. This PR does not close issue #510 by itself.

Test-suite impact
-----------------

One existing examples test (`promise/any/promise-any-failure`) had a history snapshot that encoded the buggy (lower) `InvocationCompleted` count. The history was regenerated.

One internal testing-package integration test
(`should track operations across multiple invocations`) had `expect(invocations).toHaveLength(2)`. With the fix, the natural execution of `wait → step → wait → return` produces 3 invocations; the assertion was relaxed to accept 2 or 3 (both valid depending on race with the termination cooldown).

*Issue #, if available:*

*Description of changes:*


ayushiahjolia:
Closing this stale PR. Related issue is closed already - https://github.com/aws/aws-durable-execution-sdk-js/issues/510.

ayushiahjolia:
This PR is addressing separate issue from https://github.com/aws/aws-durable-execution-sdk-js/issues/510, re-opening PR.

PR #523, opened 2026-05-01
feat(sdk): add preview config to createFileSystemSerdes

ParidelPooya:
Adds an optional generatePreview function to FileSystemSerdesConfig that stores a subset of the serialized value inline in the checkpoint envelope alongside the file pointer. This makes step data visible in the
  AWS console and API without reading the full file from the filesystem.

**New generatePreview option in FileSystemSerdesConfig:**

```
  createFileSystemSerdes("/mnt/s3", {
    generatePreview: (value) => {...}
  });
```

  Checkpoint envelope with preview:

```
  { "file": "/mnt/s3/.../1.json", "preview": { "id": "order-123", "user": { "email": "***" } } }
```

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
+++ b/packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
@@ -167,6 +167,23 @@ function isMatched(path: string, fields: PreviewField[] | undefined): boolean {
  * @internal
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
+function setNestedValue(
+  obj: Record<string, unknown>,
+  path: string,
+  value: unknown,
+): void {
+  const parts = path.split(".");
+  let current = obj;
+  for (let i = 0; i < parts.length - 1; i++) {
+    if (!(parts[i] in current) || typeof current[parts[i]] !== "object") {
+      current[parts[i]] = {};
+    }
+    current = current[parts[i]] as Record<string, unknown>;
+  }
+  current[parts[parts.length - 1]] = value;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function buildPreview(
   value: any,
   config: PreviewConfig,
@@ -204,10 +221,11 @@ export function buildPreview(
       }
 
       if (masked) {
-        const candidate = { ...preview, [path]: maskString };
+        const candidate = JSON.parse(JSON.stringify(preview));
+        setNestedValue(candidate, path, maskString);
         if (Buffer.byteLength(JSON.stringify(candidate), "utf-8") > maxBytes)
           return;
-        preview[path] = maskString;
+        setNestedValue(preview, path, maskString);
         continue;
       }
 
@@ -221,10 +239,11 @@ export function buildPreview(
         continue;
       }
 
-      const candidate = { ...preview, [path]: obj[key] };
+      const candidate = JSON.parse(JSON.stringify(preview));
+      setNestedValue(candidate, path, obj[key]);
       if (Buffer.byteLength(JSON.stringify(candidate), "utf-8") > maxBytes)
         return;
-      preview[path] = obj[key];
+      setNestedValue(preview, path, obj[key]);
     }
   }
 
--- a/packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
+++ b/packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
@@ -174,18 +174,19 @@ function setNestedValue(
 ): void {
   const DANGEROUS_KEYS = new Set(["__proto__", "constructor", "prototype"]);
   const parts = path.split(".");
+  if (parts.some((p) => DANGEROUS_KEYS.has(p))) return;
   let current = obj;
   for (let i = 0; i < parts.length - 1; i++) {
-    if (DANGEROUS_KEYS.has(parts[i])) return;
-    if (!(parts[i] in current) || typeof current[parts[i]] !== "object") {
-      current[parts[i]] = {};
+    if (
+      !Object.prototype.hasOwnProperty.call(current, parts[i]) ||
+      typeof current[parts[i]] !== "object" ||
+      current[parts[i]] === null
+    ) {
+      current[parts[i]] = Object.create(null) as Record<string, unknown>;
     }
     current = current[parts[i]] as Record<string, unknown>;
   }
-  const lastKey = parts[parts.length - 1];
-  if (!DANGEROUS_KEYS.has(lastKey)) {
-    current[lastKey] = value;
-  }
+  current[parts[parts.length - 1]] = value;
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
+++ b/packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
@@ -183,6 +183,15 @@ export function buildPreview(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   function collect(obj: any, pathPrefix: string): void {
     if (obj === null || typeof obj !== "object") return;
+
+    // Recurse into arrays transparently (indices are not part of the path)
+    if (Array.isArray(obj)) {
+      for (const item of obj) {
+        collect(item, pathPrefix);
+      }
+      return;
+    }
+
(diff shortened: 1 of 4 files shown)


packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
```diff
@@ -59,11 +128,163 @@ export interface FileSystemSerdesConfig {
  * filesystem via S3 Files, enabling durable, shared state across invocations
  * and parallel function instances without checkpoint size constraints.
  *
+/** @internal */
```
yaythomas, line 76: stray /**

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
```diff
@@ -59,11 +128,163 @@ export interface FileSystemSerdesConfig {
...
+        (masked ||
+          (config.mode === PreviewMode.INCLUDE_ALL
+            ? true
+            : isMatched(path, config.include)));
+
+      if (!visible) {
+        if (
+          obj[key] !== null &&
+          typeof obj[key] === "object" &&
+          !Array.isArray(obj[key])
```
yaythomas: does this `!Array` mean  `{ items: [{ secret: "xyz" }] }` would not mask?

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
```diff
@@ -59,11 +128,163 @@ export interface FileSystemSerdesConfig {
...
+
+  // Step 1: collect flat (path, displayValue) pairs
+  const pairs: Array<[string, unknown]> = [];
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  function collect(obj: any, pathPrefix: string): void {
+    if (obj === null || typeof obj !== "object") return;
+    for (const key of Object.keys(obj)) {
+      if (DANGEROUS_KEYS.has(key)) continue;
+      const path = pathPrefix ? `${pathPrefix}.${key}` : key;
```
yaythomas: `{ user: { email: "x" } }` and `{ "user.email": "x" }` will end up as `user.email`?

So PATH and ANYWHERE matches on either.

```
  if (mode === FieldMatchMode.PATH) {
    return path === field.name;
  }
...
 return path.split(".").includes(field.name);
```

What then happens on the rebuild?

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
```diff
@@ -59,11 +128,163 @@ export interface FileSystemSerdesConfig {
...
+
+  // Build nested structure from accepted pairs in O(n).
+  // Keys are safe at this point — dangerous keys were filtered in collect().
+  const result: Record<string, unknown> = {};
+  for (const [path, val] of accepted) {
+    const parts = path.split(".");
+    let node = result;
+    for (let i = 0; i < parts.length - 1; i++) {
+      if (typeof node[parts[i]] !== "object" || node[parts[i]] === null) {
+        node[parts[i]] = {};
```
yaythomas: if you had `result.user = "arb"`, and then later `user.email`, this ends up as `node['user'] = {}`, and you lose `arb`?

yaythomas (commented):
This is relatively complicated.

I wonder if just providing the caller with the ability to specify their own transform means the SDK can be less opinionated about this?


```ts
   export interface FileSystemSerdesConfig {
     storageMode?: FileSystemSerdesMode;
     preview?: (value: unknown) => unknown;
     maxPreviewBytes?: number;   // keep as sanity guard, default 4096
   }
```

Usage:

```ts
   createFileSystemSerdes("/mnt/s3", {
     preview: (v) => ({ id: v.id, status: v.status, email: "***" }),
   });

   // or with a utility:
   import { omit } from "lodash";
   createFileSystemSerdes("/mnt/s3", {
     preview: (v) => omit(v, ["password", "ssn"]),
   });
```


You could combine with a helper for masking:

```ts
      export const maskFields = (keys: string[], maskString = "***") =>
        (v: any) => keys.reduce((acc, k) => ({ ...acc, [k]: maskString }), v);

```

e.g

```
   maskFields(["ssn"]).
```

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
ParidelPooya: Good catch, chenged

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
ParidelPooya: Do you mean when the field name has `.` in it? I think it's acceptable to not support it. In that case customer can change the name or write their own Serdes. Updated TS Doc to be clear about it.

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
ParidelPooya: Note it can't occur with valid JSON, and optionally add a guard:
```
   if (typeof node[parts[i]] !== "object" || node[parts[i]] === null) {
     node[parts[i]] = {}; // only overwrites if not already an object — safe with valid JSON
   }
   ```

File: packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts (hunk 1 above)
Comment (ParidelPooya) [reply packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts:76]: Fixed

File: packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts [hunk 5]
```diff
@@ -97,8 +128,25 @@ export interface FileSystemSerdesConfig {
  * context.configureSerdes({
  *   defaultSerdes: createFileSystemSerdes("/mnt/s3", { storageMode: FileSystemSerdesMode.OVERFLOW }),
  * });
+ *
+ * // With preview: show id and masked email in checkpoint
+ * context.configureSerdes({
+ *   defaultSerdes: createFileSystemSerdes("/mnt/s3", {
+ *     preview: {
```
yaythomas: this isn't correct anymore?  The config field is generatePreview, and it's a function, not a PreviewConfig object.

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
```diff
@@ -2,6 +2,14 @@ import { mkdir, readFile, writeFile } from "node:fs/promises";
 import { join } from "node:path";
 import { SerdesContext, AnySerdes } from "./serdes";
 import { CHECKPOINT_SIZE_LIMIT_BYTES } from "../constants/constants";
+import { buildPreview } from "./preview";
+export {
+  FieldMatchMode,
+  PreviewMode,
+  PreviewField,
+  PreviewConfig,
+  buildPreview,
```
yaythomas, line 10: unused

packages/aws-durable-execution-sdk-js/src/utils/serdes/preview.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  const pairs: Array<[string, unknown]> = [];
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  function collect(obj: any, pathPrefix: string): void {
+    if (obj === null || typeof obj !== "object") return;
+
+    if (Array.isArray(obj)) {
+      for (const item of obj) {
+        collect(item, pathPrefix);
```
yaythomas, line 131: same prefix for every element.

what if elements are not homogeneous at the same sub-field?

```
  const value = { items: [{ user: "arb" }, { user: { email: "x" } }] };
  buildPreview(value, { mode: PreviewMode.INCLUDE_ALL });

//  "arb" lost
// => { items: { user: { email: "x" } } }  
```

this might be a necessary limitation? maybe docstring update like

> when array elements have heterogeneous shapes at the same field, later objects will overwrite earlier primitives in the preview.

ParidelPooya: Added

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
ParidelPooya: Fixed

yaythomas (approved):
:shipit:

PR #525, opened 2026-05-05
fix(sdk): preserve errorData across runInChildContext boundaries

### LINKED ISSUE #524: [Bug]: DurableOperationError.toErrorObject loses errorData across runInChildContext boundaries
### Expected Behavior

An error thrown with `errorData` set should keep that payload findable at the catch site after the throw crosses one or more `runInChildContext` boundaries — regardless of how many. The wrap-and-rethrow is performed by the SDK; application code never gets a hand on the wrapper to set `errorData` on it.

### Actual Behavior

After two boundary crossings, `errorData` is irrecoverable.

The catch path inside `runInChildContext` serialises the throw via `DurableOperationError.toErrorObject` (`dist-cjs/index.js:677-684`), which copies only `this.errorData`:

```js
toErrorObject() {
  return {
    ErrorType: this.errorType,
    ErrorMessage: this.message,
    ErrorData: this.errorData,   // single-level; deeper cause not consulted
    StackTrace: undefined,
  };
}
```

It then reconstructs via `fromErrorObject` and wraps the result in a fresh `ChildContextError(message, cause)` whose own `errorData` is left undefined (the constructor takes `(message, cause)` only). Each crossing therefore consumes one `errorData` position on the chain. After the second crossing the deeper value has been overwritten by another undefined and is gone.

### Steps to Reproduce

The following snippet simulates the SDK's own catch path inline (matching `dist-cjs/index.js` ~line 1673), so it's deterministic and harness-free:

```ts
import {
  CallbackError,
  ChildContextError,
  DurableOperationError,
} from '@aws/durable-execution-sdk-js';

const simulateChildContextCatch = (err: unknown): ChildContextError => {
  const obj = err instanceof DurableOperationError
    ? err.toErrorObject()
    : { ErrorMessage: (err as Error).message, ErrorType: (err as Error).name };
  const reconstructed = DurableOperationError.fromErrorObject(obj);
  return new ChildContextError(reconstructed.message, reconstructed);
};

const SENTINEL = JSON.stringify({ reason: 'operator-cancelled' });
const leaf = new CallbackError('cb failed', undefined, SENTINEL);

let outer: unknown = leaf;
for (let i = 0; i < 2; i++) outer = simulateChildContextCatch(outer);

let node: unknown = outer;
let found: string | undefined;
for (let i = 0; i < 10 && node; i++) {
  if (typeof node === 'object' && typeof (node as any).errorData === 'string') {
    found = (node as any).errorData;
    break;
  }
  node = (node as any).cause;
}
console.log(found); // expected: SENTINEL — actual: undefined
```

After 1 wrap the `errorData` survives on `outer.cause`. After 2, it is gone.

### SDK Version

1.1.1

### Node.js Version

22.x

### Is this a regression?

No

### Last Working Version

_No response_

### Additional Context

Also reproduced on `1.1.1`; `dist-cjs/index.js:677-684` is byte-identical between 1.1.1 and 1.1.2, so the bug is structural to the wrap path.

**Proposed fix** — make `toErrorObject` chain-aware: when its own `errorData` is undefined, walk `this.cause` (bounded — same depth as `fromErrorObject`'s reconstruction loop) and surface the first `errorData: string` it finds:

```ts
toErrorObject() {
  let errorData = this.errorData;
  if (errorData === undefined) {
    let node: unknown = this.cause;
    for (let i = 0; i < 10 && node; i++) {
      if (typeof node === 'object' && typeof (node as any).errorData === 'string') {
        errorData = (node as any).errorData;
        break;
      }
      node = (node as any).cause;
    }
  }
  return {
    ErrorType: this.errorType,
    ErrorMessage: this.message,
    ErrorData: errorData,
    StackTrace: undefined,
  };
}
```

Self-contained ~10-line change. Doesn't touch the wrap path; doesn't change the public type; doesn't add new API.

**Why it matters in practice** — we use typed `errorData` payloads to encode operator-cancel sentinels and typed provider-failure classifications. Across our orchestrator's nested parallel/coordinator structure, a thrown error commonly crosses 3–4 `runInChildContext` boundaries between source and the terminal handler, so the deeper-than-2 case is routine. Today we work around it with a manual lift at every helper that catches and re-throws across a boundary, mutating the throw in place before the SDK's catch sees it:

```ts
const liftErrorData = (err: unknown): void => {
  if (typeof err !== 'object' || err === null) return;
  if (typeof (err as any).errorType !== 'string') return;
  if (typeof (err as any).errorData === 'string') return;
  let node: unknown = (err as any).cause;
  for (let i = 0; i < 10 && node; i++) {
    if (typeof node === 'object' && typeof (node as any).errorData === 'string') {
      (err as any).errorData = (node as any).errorData;
      return;
    }
    node = (node as any).cause;
  }
};
```

Happy to open a PR with the fix + a unit test if helpful.

t-d-d:
When `runInChildContext` re-wraps a replayed `DurableOperationError`, the handler did `new ChildContextError(originalError.message, originalError)` without forwarding `errorData`, so the typed payload disappeared after the second boundary crossing -- breaking error-classification flows that depend on it (operator-cancellation signals, failure categories, etc.).

Forward `originalError.errorData` at both wrap sites (`run-in-child-context-handler.ts:224,435`). The wrapper now carries the payload in-memory and on the wire, all the way up the boundary stack.

User-supplied `errorMapper` callbacks must forward `errorData` themselves if they want it preserved -- e.g.
`errorMapper: (e) => new MyError(e.message, e, e.errorData)`.

Fixes #524


ParidelPooya:
Addressed in this commit: https://github.com/aws/aws-durable-execution-sdk-js/commit/8abeada650cdde9bea4b5f60f69beef960a338aa

PR #526, opened 2026-05-05
fix(sdk): always re-evaluate termination after checkpoint queue drains

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-js/issues/510

*Description of changes:*
Removes the `if (!this.terminationTimer)` guard in processQueue's finally block so that termination is always re-evaluated after the checkpoint queue drains.

### Problem

In map + waitForCallback scenarios, callbacks register as IDLE_AWAITED and set terminationTimer before a completed branch's fire-and-forget checkpoint enters the queue. When the queue drains, the guard sees terminationTimer is
non-null and skips checkAndTerminate. The function never suspends - it wedges until an unrelated timer (e.g., a sibling ctx.wait) fires.

### Fix

Replace the conditional with an unconditional `this.checkAndTerminate()`.
This is safe because `scheduleTermination` already returns early when the same reason is already scheduled — no timer restart, no extra cooldown in the common case.

### Testing

- Added unit tests that fail without the fix and pass with it
- Added map + waitForCallback and map + waitForCondition example tests
- All existing tests pass 

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/src/examples/map/with-callback/map-with-callback-callback-failure.history.json
+++ b/packages/aws-durable-execution-sdk-js-examples/src/examples/map/with-callback/map-with-callback-callback-failure.history.json
@@ -1,0 +1,354 @@
+[
+  {
+    "EventType": "ExecutionStarted",
+    "EventId": 1,
+    "Id": "82934803-886d-4488-a8e6-120038632717",
+    "EventTimestamp": "2026-05-01T19:30:08.193Z",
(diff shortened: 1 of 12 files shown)


ParidelPooya (changes requested):
I ran new added examples and integration test with and without the fix (change to checkpoint-manager), ignoring assertEventSignatures, integration test are passing before and after the change.

That means that we didn't manage to reproduce the customer issue.

PR #527, opened 2026-05-05
feat(sdk): Lambda bundled runtime useragent header

nvasiu:
*Issue #, if available:*

N/A

*Description of changes:*

**Note:** we are waiting to confirm the Lambda bundled runtime SDK path. Do NOT merge this PR until we get confirmation.

Updating the UserAgent header string to indicate whether the SDK is being run within the Lambda bundled runtime.

- A new boolean checks if the SDK's current path matches the path where it would be installed in the Lambda bundled runtime.
- If the above boolean is true, the string inserted into the UserAgent header will include "-bundled". Example:
`aws-durable-execution-sdk-js/{version}bundled`

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/utils/constants/version.ts
+++ b/packages/aws-durable-execution-sdk-js/src/utils/constants/version.ts
@@ -7,11 +7,19 @@
  *
  * SDK_NAME is a fixed string matching the cross-SDK convention:
  * aws-durable-execution-sdk-\{language\}
+ * Alternate version if SDK path is within the Lambda bundled runtime.
  *
  * Defaults are provided for test environments where Rollup doesn't run
  * and process.env values are undefined.
  *
  * @internal
  */
+const runtimeDir = process.env.LAMBDA_RUNTIME_DIR || "/var/runtime";
+const isRuntimeBundled =
+  typeof __dirname !== "undefined" && __dirname.startsWith(runtimeDir);
+
 export const SDK_NAME = "aws-durable-execution-sdk-js";
-export const SDK_VERSION = process.env.NPM_PACKAGE_VERSION || "0.0.0";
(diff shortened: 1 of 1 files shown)


packages/aws-durable-execution-sdk-js/src/utils/constants/version.ts
```diff
@@ -7,11 +7,16 @@
...
  * aws-durable-execution-sdk-\{language\}
+ * Alternate version if SDK path is within the Lambda bundled runtime.
  *
  * Defaults are provided for test environments where Rollup doesn't run
  * and process.env values are undefined.
  *
  * @internal
  */
-export const SDK_NAME = "aws-durable-execution-sdk-js";
+const isRuntimeBundled = __dirname.startsWith("/var/runtime");
```
ParidelPooya: ES Module Compatibility:  __dirname doesn't exist in ES Modules. By switching  "type": "module", this code will throw a crash-inducing ReferenceError.

Something like this code is safer:
```
// Using official environment variables is more reliable
const runtimeDir = process.env.LAMBDA_RUNTIME_DIR || "";

const currentDir = typeof __dirname !== 'undefined' 
  ? __dirname 
  : import.meta.dirname;

const isRuntimeBundled = currentDir?.startsWith(process.env.LAMBDA_RUNTIME_DIR || "/var/runtime");
```

ParidelPooya (changes requested):

packages/aws-durable-execution-sdk-js/src/utils/constants/version.ts
nvasiu: `import.meta.dirname` fails with CJS.

Updated the PR to instead just check if __dirname exists, and if not just default to the non-runtime header.

Also the update uses the official env var.

bchampp (approved):

ParidelPooya (approved):

PR #544, opened 2026-05-19
  fix(testing): keep TimerScheduler entries in flight until startInvocation settles

joeyzhao2018:
TimerScheduler.scheduleFunction used to remove the timer from runningTimers synchronously the moment the setTimeout fired — before the chained updateCheckpoint().then(startInvocation) had a chance to begin, let alone finish. That left a window where the underlying scheduled work was running but hasScheduledFunction() returned false.

The window is load-bearing for TestExecutionOrchestrator's invocation- completion validation (test-execution-orchestrator.ts:651-663):

  if (!this.scheduler.hasScheduledFunction() && !this.pendingOperations.size) {
    // reject with "Cannot return PENDING status with no pending operations."
  }

For fast-resolving chained-invoke flows — and for waitForCallback flows where the test harness resolves the callback immediately — the resume timer can fire and tear down its runningTimers entry between the SDK's PENDING-returning invokeHandler completing and the validation running. The outer's pendingOperations slot is already cleared by then, so both conditions report empty and the validation throws.

In production (real AWS backend) this race doesn't manifest because the backend tracks the actual operation graph the SDK reports through checkpoint updates. The validation is mock-only.

Fix: move runningTimers.delete(timer) out of the synchronous setTimeout callback into a .finally(...) chained on the updateCheckpoint -> startInvocation promise, so a scheduled function stays counted from the moment it's scheduled until its work settles (success or failure). Updates the JSDoc on hasScheduledFunction() to describe the new "pending OR in-flight" semantics — the previous TODO comment ("if there is no scheduled function and the invocation status is PENDING, the language SDK has a bug") was contradicted by the fix and would mislead future maintainers. Adds two regression tests covering the in-flight success and rejection paths.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/aws-durable-execution-sdk-js-testing/src/test-runner/local/orchestration/__tests__/timer-scheduler.test.ts
+++ b/packages/aws-durable-execution-sdk-js-testing/src/test-runner/local/orchestration/__tests__/timer-scheduler.test.ts
@@ -171,6 +171,44 @@ describe("TimerScheduler", () => {
       expect(scheduler.hasScheduledFunction()).toBe(false);
     });
 
+    it("should keep hasScheduledFunction true while the fired function is still running", async () => {
+      let resolveInvocation: (() => void) | undefined;
+      const invocationPromise = new Promise<void>((resolve) => {
+        resolveInvocation = resolve;
+      });
+      const mockFn = jest.fn().mockReturnValue(invocationPromise);
+      const mockOnError = jest.fn();
+
+      scheduler.scheduleFunction(mockFn, mockOnError);
+
+      await jest.advanceTimersByTimeAsync(0);
+      expect(mockFn).toHaveBeenCalledTimes(1);
+      expect(scheduler.hasScheduledFunction()).toBe(true);
+
+      resolveInvocation!();
+      await jest.advanceTimersByTimeAsync(0);
+      expect(scheduler.hasScheduledFunction()).toBe(false);
+    });
+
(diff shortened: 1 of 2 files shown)


anthonyting:
Could you provide an example of a test for this case which fails before and succeeds now? Ideally we can also add an integration test for this case following the instructions in https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js-examples/ADDING_EXAMPLES.md

ayushiahjolia:
Great bug find - the race condition analysis is correct. However, the `.finally()` placement over-corrects.

Since startInvocation is invokeHandler(), keeping the timer alive through its entire execution means the PENDING validation at the end of invokeHandler:
```
if (!this.scheduler.hasScheduledFunction() && !this.pendingOperations.size) {
  // reject: "Cannot return PENDING status with no pending operations."
}
```
will always see `hasScheduledFunction() === true` (we're still inside the chain, .finally() hasn't run). This makes the check dead code. Running the existing tests confirms this, 3 tests timeout with this fix.

The race window is only between timer firing and updateCheckpoint() completing. Deleting the timer after checkpoint but before invocation fixes it without breaking the validation.
  
I've opened a PR for this - https://github.com/aws/aws-durable-execution-sdk-js/pull/611

Before we merge my PR, please provide the example of integration test as Anthony said, where you encountered this issue.

ParidelPooya:
Please re-open this with an Integration test that shows what is the current issue and how the change is fixing the issue.

Issue #543, opened 2026-05-18
[Bug]: v1.1.3 breaks when bundled to CJS - Regression from 1.1.2


zahir-haast:
### Expected Behavior

Version 1.1.3 should work when bundled into CJS by esbuild as it did in 1.1.2.   I'm using it with the default AWS CDK Lambda which by default uses esbuild with cjs by default.  

### Actual Behavior

A runtime error is thrown immediately on import:

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of URL. Received undefined

Root cause

Commit 2655eb0 added a __dirname reference, but since it doesn't ESM, Rollup seems to injects a CJS compatibility shim at the top of dist/index.mjs:

// -- Shims --
import cjsUrl from 'node:url';
import cjsPath from 'node:path';
import cjsModule from 'node:module';
const __filename = cjsUrl.fileURLToPath(import.meta.url);
const __dirname = cjsPath.dirname(__filename);
const require = cjsModule.createRequire(import.meta.url);

You can see this if you diff 1.1.2 vs 1.1.3 index.mjs outputs

This shim works when consumed as native ESM, but when a downstream bundler re-bundles the .mjs into CJS format, import.meta is replaced with an empty object {}, cascading the failure

### Steps to Reproduce

Install 1.1.3
Bundle some hello world code with esbuild: `esbuild --bundle --format=cjs --platform=node`
Run the output - it'll throw an error.

### SDK Version

1.1.3

### Node.js Version

22.x

### Is this a regression?

Yes

### Last Working Version

1.1.2

PR #546, opened 2026-05-20
fix(sdk): resolve CJS bundling regression in v1.1.3

ParidelPooya:
When the SDK is bundled to CommonJS format using esbuild, it throws TypeError [ERR_INVALID_ARG_TYPE]: The "url" argument must be of type string. Received undefined due to Rollup's ESM shim converting import.meta.url to {} in CJS bundles.

  Solution

  - Replace direct import.meta access with Function constructor approach to avoid Rollup's ESM shim interference
  - Add Jest compatibility with early return for test environments

  Testing

  - All 75 test suites pass (842/842 tests)
  - ESM and CJS builds successful
  - esbuild CJS bundling now works (original issue resolved)
  - Jest compatibility maintained
  
*Issue #, if available:*
#543


yaythomas (dismissed):

packages/aws-durable-execution-sdk-js/rollup.config.mjs
```diff
@@ -3,12 +3,41 @@
 import { createBuildOptions } from "../../rollup.config.js";
 import packageJson from "./package.json" with { type: "json" };
 
+// Custom ESM shim that's safer when bundled to CJS
```
anthonyting: should the other esm shim be removed (and replaced with this one, since testing lib also needs an esm shim)? https://github.com/aws/aws-durable-execution-sdk-js/blob/6d57ea2c57729967865b6f267f61fc4e86929077/rollup.config.js#L75

yaythomas (approved):

Issue #545, opened 2026-05-20
[Feature]: Plugin hooks interface and wiring the hooks into handlers


SilanHe:
### What would you like?

Instead of implementing observability directly, the SDK exposes a lightweight instrumentation plugin interface. Plugins are:

- Optional — zero cost if unused, no dependencies added to the bundle
- Composable — multiple plugins can be registered and all receive every hook call
- General-purpose — the same interface supports tracing, metrics, cost tracking, execution summaries, audit logs, or anything else

We need something that is sufficient for composing an Otel SDK at the minimum. We'd like something generic enough that it can be used for other plugin use cases.

### Possible Implementation

This could be implemented according to https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/documents/INSTRUMENTATION_PLUGIN.md

### Is this a breaking change?

No

### Does this require an RFC?

No

PR #548, opened 2026-05-20
feat(sdk): Plugin interface and runner

SilanHe:
*Issue #, if available:* #545

*Description of changes:* 

Introduce the plugin system foundation for the durable execution SDK:

- DurableInstrumentationPlugin interface with lifecycle hooks for
  execution, invocation, operation, and attempt-level events
- Plugin runner that composes multiple plugins with fire-and-forget and callback-wrapping execution semantics
- Operation utility helpers (toOperationInfo, toAttemptInfo, etc.)
- Export plugin types from the public API
- Add plugins config field to DurableExecutionConfig

This is a pure additive change with no wiring into the SDK handlers.
Subsequent PRs will wire the hooks into the execution lifecycle and
individual operation handlers.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/types/plugin.ts
+++ b/packages/aws-durable-execution-sdk-js/src/types/plugin.ts
@@ -1,5 +1,10 @@
 import { Operation } from "@aws-sdk/client-lambda";
 
+/**
+ * Information about a durable operation.
+ *
+ * @experimental This function is experimental and may be changed or removed in future releases.
+ */
 export interface OperationInfo {
   Id: string;
   Name?: string;
@@ -10,16 +15,31 @@ export interface OperationInfo {
   EndTimestamp?: Date;
 }
 
+/**
+ * Information about an operation attempt.
+ *
+ * @experimental This function is experimental and may be changed or removed in future releases.
+ */
 export interface AttemptInfo extends OperationInfo {
   Attempt: number;
 }
 
+/**
+ * Possible outcomes for an operation attempt.
+ *
+ * @experimental This function is experimental and may be changed or removed in future releases.
+ */
 export enum AttemptEndInfoOutcome {
   SUCCEEDED = "succeeded",
   FAILED = "failed",
   RETRYING = "retrying",
 }
 
+/**
+ * Information provided when an operation attempt ends.
+ *
+ * @experimental This function is experimental and may be changed or removed in future releases.
+ */
 export interface AttemptEndInfo extends AttemptInfo {
   outcome:
     | AttemptEndInfoOutcome.SUCCEEDED
@@ -29,11 +49,21 @@ export interface AttemptEndInfo extends AttemptInfo {
   nextAttemptDelaySeconds?: number;
 }
 
+/**
+ * Information about a durable execution invocation.
+ *
+ * @experimental This function is experimental and may be changed or removed in future releases.
+ */
 export interface InvocationInfo {
   requestId: string;
   executionArn: string;
 }
 
+/**
+ * Information provided when a durable execution ends.
+ *
+ * @experimental This function is experimental and may be changed or removed in future releases.
+ */
 export interface ExecutionEndInfo extends InvocationInfo {
   status: "SUCCEEDED" | "FAILED";
   executionResult?: unknown;
@@ -42,11 +72,21 @@ export interface ExecutionEndInfo extends InvocationInfo {
   operations: Record<string, Operation>;
 }
 
+/**
+ * Information provided when operations change during execution.
+ *
+ * @experimental This function is experimental and may be changed or removed in future releases.
+ */
 export interface OperationChangeInfo extends InvocationInfo {
   updatedOperations: Record<string, Operation>;
   operations: Record<string, Operation>;
 }
 
+/**
+ * Plugin interface for instrumenting durable execution lifecycle events.
+ *
+ * @experimental This function is experimental and may be changed or removed in future releases.
+ */
 export interface DurableInstrumentationPlugin {
   onExecutionStart?(info: InvocationInfo): void;
   onExecutionEnd?(info: ExecutionEndInfo): void;
--- a/packages/aws-durable-execution-sdk-js/src/types/plugin.ts
+++ b/packages/aws-durable-execution-sdk-js/src/types/plugin.ts
@@ -1,0 +1,79 @@
+import { Operation } from "@aws-sdk/client-lambda";
+
+export interface OperationInfo {
+  Id: string;
+  Name?: string;
+  Type: string;
+  SubType?: string;
+  ParentId?: string;
+  StartTimestamp?: Date;
+  EndTimestamp?: Date;
+}
+
+export interface AttemptInfo extends OperationInfo {
+  Attempt: number;
+}
+
+export enum AttemptEndInfoOutcome {
+  SUCCEEDED = "succeeded",
+  FAILED = "failed",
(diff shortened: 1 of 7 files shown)


packages/aws-durable-execution-sdk-js/src/utils/operation/operation.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Backfills missing fields on an OperationInfo (or subtype) with the provided defaults.
+ * Only sets a field if it's not already present (undefined or empty string).
+ *
+ * @experimental This function is experimental and may be changed or removed in future releases.
+ */
+export function backfillOperationInfo<T extends OperationInfo>(
+  info: T,
+  defaults: Partial<OperationInfo>,
+): T {
+  info.Id = defaults.Id ?? "";
```
ayushiahjolia, line 78: why are we overriding this always?

ayushiahjolia (dismissed):

packages/aws-durable-execution-sdk-js/src/utils/operation/operation.ts
SilanHe: it's because the Id we get back from getStepData is different from the raw Id, it's hashed once it's returned from getStepData. I'm trying to always provide a consistent Id and I chose to provide the unhashed Id.

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+  nextAttemptDelaySeconds?: number;
+}
+
+/**
+ * Information about a durable execution invocation.
+ *
+ * @experimental This interface is experimental and may be changed or removed in future releases.
+ */
+export interface InvocationInfo {
+  requestId: string;
```
ParidelPooya, line 69: Can we add another field here to show it's the initial invocation or not? 
If possible, then please create a GitHub issue for it.

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+  executionArn: string;
+}
+
+/**
+ * Information provided when a durable execution ends.
+ *
+ * @experimental This interface is experimental and may be changed or removed in future releases.
+ */
+export interface ExecutionEndInfo extends InvocationInfo {
+  status: "SUCCEEDED" | "FAILED";
```
ParidelPooya, line 80: Should we add  another one for STOPPED and TIMEOUT (execution timeout) I can't remember how SDK right now deal those status but those are possible (we call checkpoints and checkpoint will throw an error that shows that the execution is stopped or timeedout)

ParidelPooya (commented):
I see multiple instance of usage for `any`, can we use a better type? If not we can still create a type that maps to `any` but the code becomes more readable.
For example if we use `any` for operationResult (an example, I'm not saying we use result in here) then let's define `operationResult` and then use it instead of `any`

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
SilanHe: yes

SilanHe: if I add that field, I will remove onExecutionStart hook as the new field makes it redundant.

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
SilanHe: Hmm I thought that we didn't know whether an execution was STOPPED or TIMEOUT. Looking at the public docs for the CheckpointDurableExecution and GetDurableExecutionState, it seems that we would instead receive an InvalidParameterValueException due to invalid checkpoint token. In this case, I think what happens is that the this invocation will fail (maybe after some retries for the checkpoint). The backend will see that the execution is STOPPED or TIMEOUT and the lambda won't be re-invoked.

### artifacts
* https://docs.aws.amazon.com/lambda/latest/api/API_CheckpointDurableExecution.html
* https://docs.aws.amazon.com/lambda/latest/api/API_GetDurableExecutionState.html
* where we process the InvalidParameterValueException: https://github.com/aws/aws-durable-execution-sdk-js/blob/28c2d42ee5ebd5c72ad9cff5bff5f2dd5403c662/packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts#L235
* Types we do seem to handle in SDK: https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/types/core.ts#L89

SilanHe:
are you talking about the any in `createPluginRunner`?

The any from [line 60](https://github.com/aws/aws-durable-execution-sdk-js/blob/bcea797ed2ba70949b1b755f6b92064705ad2cb3/packages/aws-durable-execution-sdk-js/src/utils/plugin/plugin-runner.ts#L60) covers the result/exception from the customer defined fn. I'll replace any with CustomerFnResult;

The any from [line 47](https://github.com/aws/aws-durable-execution-sdk-js/blob/bcea797ed2ba70949b1b755f6b92064705ad2cb3/packages/aws-durable-execution-sdk-js/src/utils/plugin/plugin-runner.ts#L47) represents a callback with 2 parameters. I could type this but then the type would also be comprised of any...

[line 77](https://github.com/aws/aws-durable-execution-sdk-js/blob/bcea797ed2ba70949b1b755f6b92064705ad2cb3/packages/aws-durable-execution-sdk-js/src/utils/plugin/plugin-runner.ts#L77) typically represent either a callback function with 1 variable.

I'll introduce the following:

```
  type CustomerFnResult = any;
  type CallbackResult = any;
  type CustomerFn = () => CustomerFnResult;
  type PluginInfo =
    | OperationInfo
    | InvocationInfo
    | ExecutionEndInfo
    | AttemptEndInfo
    | AttemptInfo
    | OperationChangeInfo
    | undefined;
  type PluginWrapperHookFn = (
    info: PluginInfo,
    fn: CustomerFnResult,
  ) => CallbackResult;
  type PluginHookFn = (info: PluginInfo) => CallbackResult;
```

packages/aws-durable-execution-sdk-js/src/utils/plugin/plugin-runner.ts
```diff
@@ -0,0 +1,153 @@
...
+    onOperationAttemptStart: (info: AttemptInfo) =>
+      run("onOperationAttemptStart", info),
+    wrapOperationAttemptFn: (
+      info: AttemptInfo,
+      fn: CustomerFn,
+    ): CustomerFnResult => runAsCallback("wrapOperationAttemptFn", info, fn),
+    onOperationAttemptEnd: (info: AttemptEndInfo) =>
+      run("onOperationAttemptEnd", info),
+    onOperationChange: (info: OperationChangeInfo) =>
+      run("onOperationChange", info),
```
zhongkechen, line 140: Looks like the APIs are more than required.

zhongkechen (approved):

ayushiahjolia (approved):

packages/aws-durable-execution-sdk-js/src/utils/plugin/plugin-runner.ts
SilanHe: onOperationChange you mean?

zhongkechen: The entire interface in general

PR #551, opened 2026-05-21
ci: Opt into Node.js 24 for GitHub Actions runners

ayushiahjolia:
*Issue #, if available:* N/A

*Description of changes:*
## Summary

Add `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true` environment variable to all workflows that use `actions/checkout@v4` and `actions/setup-node@v4`.

GitHub is deprecating Node.js 20 for Actions runners. Node.js 24 will become the default on June 2nd, 2026, and Node.js 20 will be removed entirely on September 16th, 2026. This change opts in early to avoid disruption.

## Changes

Added the env variable to:
- build.yml
- capacity-provider-tests.yml
- cleanup-integration-tests.yml
- integration-tests.yml
- npm-publish.yml
- scorecard.yml
- sync-package.yml
- unit-tests.yml

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,8 @@
 name: Build
 
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
 on:
   push:
     branches: ["main", "v1.x"]
--- a/.github/workflows/capacity-provider-tests.yml
+++ b/.github/workflows/capacity-provider-tests.yml
@@ -1,5 +1,8 @@
 name: Capacity Provider Tests
 
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
 on:
   workflow_call:
     inputs:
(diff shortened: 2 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -1,5 +1,8 @@
 name: Build
 
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
```
SilanHe, line 4: hmm what's the benefit of this versus updating the version on the actions themselves?

ayushiahjolia: The FORCE_JAVASCRIPT_ACTIONS_TO_NODE24 env var is the recommended approach from GitHub for opting into Node.js 24 before the actions themselves ship new major versions that default to it. The current actions/checkout@v4 and actions/setup-node@v4 still bundle Node.js 20 as their runtime - there's no v5 yet that uses Node.js 24 natively.

This env var tells the GitHub Actions runner to use Node.js 24 to execute the action's JavaScript regardless of what the action bundles. It lets us validate our workflows against Node.js 24 early without waiting for upstream action releases. Once v5 (or equivalent) ships with Node.js 24 as default, we can drop the env var and bump the action versions instead.

SilanHe (approved):

PR #558, opened 2026-05-22
feat(sdk): add in safe DurableContext detection

willfarrell:
## Summary

Adds two class-side markers on `DurableContextImpl` so external libraries can detect a durable context without importing the SDK or relying on brittle name checks.

- **`Symbol.for("@aws/durable-execution-sdk-js/durable-context")` brand:** primary identity marker. Package-namespaced global registry key so callers can write `ctx?.[Symbol.for("@aws/durable-execution-sdk-js/durable-context")] === true` with no SDK dependency. Follows the React `$$typeof` / Vue / Apollo / Zod convention for cross-library nominal identity.
- **`Symbol.toStringTag` getter returning `"DurableContext"`:** debug/display ergonomics. `Object.prototype.toString.call(ctx)` now returns `"[object DurableContext]"`. Matches the existing pattern on `DurablePromise` and `DurableExecutionInvocationInput`.

### Motivation

Lambda middleware libraries (Powertools, Middy, etc.) need to branch behaviour on durable-vs-regular contexts. Today the options are:

  | Approach | Issue |
  |---|---|
  | `ctx.constructor.name === "DurableContextImpl"` | Mangled to `"n"` by default esbuild `--minify` |
  | `ctx instanceof DurableContextImpl` | Requires importing the SDK as a hard dep; breaks across realms or duplicate copies |
  | Duck-type `typeof ctx.step === "function"` of `'step' in ctx` | Works (what Middy / Powertools does today), but coupled to public method names |

A `Symbol.for()` brand sidesteps all three: minification-proof (string literals are not mangled), no SDK import needed, cross-realm safe via the global symbol registry, and purpose-built for nominal identity rather than display or shape.

cc @dreamorosi

ref:
- https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/idempotency/src/makeIdempotent.ts#L25

AI disclosure; used Claude Opus 4.7 to make PR. Edited and reviewed by human.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/context/durable-context/durable-context.ts
+++ b/packages/aws-durable-execution-sdk-js/src/context/durable-context/durable-context.ts
@@ -69,9 +69,15 @@ export interface DurableExecution {
   setTerminating(): void;
 }
 
+export const DURABLE_CONTEXT_BRAND = Symbol.for(
+  "@aws/durable-execution-sdk-js/durable-context",
+);
+
 export class DurableContextImpl<
   Logger extends DurableLogger,
 > implements DurableContext<Logger> {
+  readonly [DURABLE_CONTEXT_BRAND] = true;
+
   private _stepPrefix?: string;
   private _stepCounter: number = 0;
   private durableLogger: Logger;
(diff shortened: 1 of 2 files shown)


dreamorosi:
Hi @willfarrell - thanks for the PR - I'll see if I can surface it to the maintainers.

ParidelPooya (approved):

bchampp (approved):

Issue #529, opened 2026-05-08
[Bug]: ExecutionFailed event shows internal error "metadata required on first call" instead of StepInterruptedError


TrickSumo:
### Expected Behavior

When using `AtMostOncePerRetry` + `retryStrategy: () => ({ shouldRetry: false })`, the execution behaves correctly, but the final `ExecutionFailed` event in the execution history shows an internal error message instead of the expected `StepInterruptedError`.

<img width="1286" height="787" alt="Image" src="https://github.com/user-attachments/assets/d8c44c4c-4991-4752-821d-c5593bda3331" />

### Actual Behavior

The error message should be clearer.

### Steps to Reproduce

1. Create a durable Lambda function with execution timeout of 1 minute
2. Set Lambda invocation timeout to 3 seconds
3. Use a step with `AtMostOncePerRetry` + `retryStrategy: () => ({ shouldRetry: false })`
4. Inside the step, run an operation that takes longer than 3 seconds (e.g. `setTimeout(10000)`)
5. Invoke the function and observe the execution history

## Code

```typescript
import { withDurableExecution, StepSemantics } from '@aws/durable-execution-sdk-js';

export const handler = withDurableExecution(async (event, context) => {
  await context.step('Step #1', (stepCtx) => {
    stepCtx.logger.info('Hello from step #1');
  });

  await context.wait({ seconds: 1 });

  const message = await context.step('Step #2', async () => {
    context.logger.info('inside step 2');
    await new Promise(resolve => setTimeout(resolve, 10000)); // always times out
    return 'done';
  }, {
    semantics: StepSemantics.AtMostOncePerRetry,
    retryStrategy: () => ({ shouldRetry: false })
  });
});
```

### SDK Version

1.1.2

### Node.js Version

22.x

### Is this a regression?

No

PR #569, opened 2026-05-23
fix(sdk): fix AtMostOncePerRetry shouldRetry:false crashing with missing metadata and throwing wrong error type

TrickSumo:
Solves https://github.com/aws/aws-durable-execution-sdk-js/issues/529
Also aligns with the behaviour according to the documentation: https://docs.aws.amazon.com/durable-execution/sdk-reference/error-handling/errors/#step-interrupted

# Issue
When a step configured with `AtMostOncePerRetry` semantics is interrupted (e.g. Lambda timeout) and the retry strategy returns `shouldRetry: false`, two bugs occur on the next replay:

**Bug 1: crash before throw (metadata missing)**
`markOperationState` is called without `metadata` on the `shouldRetry: false` path. On replay, the checkpoint manager's `operations` map is empty (fresh Lambda instance), so the first call for that stepId has no metadata → throws: `"metadata required on first call for <stepId>"`

The `shouldRetry: true` path correctly passes metadata. There things work properly.

**Bug 2: wrong error type thrown to handler**
After fixing Bug 1, the handler receives `StepError` instead of `StepInterruptedError` because the throw goes through `DurableOperationError.fromErrorObject()` which has no case for `"StepInterruptedError"` in its switch statement → falls to default → `StepError`.

The docs state `StepInterruptedError` is thrown: https://github.com/aws/aws-durable-execution-docs/blob/90cc9acb004c75085346722c36f2528fe1d3b11c/examples/typescript/sdk-reference/error-handling/step-interrupted.ts#L7

# Description of changes:

- `step-handler.ts`: add missing `metadata` to `markOperationState` call on `shouldRetry: false` path (matching the `shouldRetry: true` path at line 201)
- `step-handler.ts`: replace `throw DurableOperationError.fromErrorObject(...)` with `throw error` to surface `StepInterruptedError` directly to the handler

Before fix:-
image

After fix:-
image

Code:-
```import { withDurableExecution, StepSemantics } from '@aws/durable-execution-sdk-js';

export const handler = withDurableExecution(async (event, context) => {

    const message = await context.step('Step1', async () => {
    await new Promise(resolve => setTimeout(resolve, 10000)); // 10 seconds
    return 'Hello from Durable Lambda!';
  }, {
    semantics: StepSemantics.AtMostOncePerRetry,
    retryStrategy: () => ({ shouldRetry: false })
  });

  const response = {
    statusCode: 200,
    body: JSON.stringify(message),
  };
  return response;
});
```

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/handlers/step-handler/step-handler.ts
+++ b/packages/aws-durable-execution-sdk-js/src/handlers/step-handler/step-handler.ts
@@ -177,10 +177,17 @@ export const createStepHandler = <Logger extends DurableLogger>(
           checkpoint.markOperationState(
             stepId,
             OperationLifecycleState.COMPLETED,
+            {
+              metadata: {
+                stepId,
+                name,
+                type: OperationType.STEP,
+                subType: OperationSubType.STEP,
+                parentId,
+              },
+            },
           );
-          throw DurableOperationError.fromErrorObject(
-            createErrorObjectFromError(error),
-          );
+          throw error;
         }
 
         await checkpoint.checkpoint(stepId, {


TrickSumo:
I'm not exactly sure if throwing StepIntruptedError directly here is a good idea: https://github.com/aws/aws-durable-execution-sdk-js/pull/569/changes#diff-1a6cc115ead41370d355110871eaefea0863a2ad5f818ad81b2d9cbcefaa632cR190

If a better way is available, please suggest.

Thanks 😊

ParidelPooya:
> I'm not exactly sure if throwing StepIntruptedError directly here is a good idea: https://github.com/aws/aws-durable-execution-sdk-js/pull/569/changes#diff-1a6cc115ead41370d355110871eaefea0863a2ad5f818ad81b2d9cbcefaa632cR190
> 
> If a better way is available, please suggest.
> 
> Thanks 😊

Hi @TrickSumo — thank you for the thorough investigation and for catching this!
  The reproduction with AtMostOncePerRetry + shouldRetry: false is exactly right, and
  the metadata fix is correct.

On your direct question ("not sure if throwing StepInterruptedError directly here
  is a good idea"): we'd suggest keeping the existing
  DurableOperationError.fromErrorObject(...) wrapping (which produces a StepError).
  The reason is the public-facing contract of step error handling:

- StepInterruptedError is currently an internal sentinel — it extends Error, not
  DurableOperationError, and it's only constructed inside step-handler.ts to be passed
  as the input argument to retryStrategy(error, attempt). So users see it inside their
  retryStrategy callback, but never as a thrown error from await context.step(...).
- Every other failure path in step-handler.ts routes the user's error through
  DurableOperationError.fromErrorObject(createErrorObjectFromError(error)), which
  produces a StepError whose cause.name preserves the original error type. That's the
  convention the existing step-error-determinism example
  (https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-
  execution-sdk-js-examples/src/examples/step/step-error-determinism/step-error-determ
  inism.ts) relies on (handlers always catch a StepError).
- The interruption signal is still preserved on the cause chain, so users can
  detect it via (err as StepError).cause?.name === "StepInterruptedError" if they need
  to differentiate.


The docs page you linked may be slightly out of sync with the code on this — happy
  to file a docs fix once we settle on the SDK behavior here.

I've put together a follow-up PR:
- Keeps your metadata fix.
- Keeps the DurableOperationError wrapping (so handlers receive StepError, with
  cause.name === "StepInterruptedError").
- Adds 3 unit-level regression tests in step-handler.test.ts covering the metadata
  contract, the error-type contract, and the retryStrategy input contract.
- Adds a cloud-only example step/interrupted-no-retry/ that reproduces the real
  Lambda-timeout interruption end-to-end (deployed with Timeout: 5s, step sleeps 30s).

Thanks again for the careful write-up and the screenshots — they made it really
  easy to verify the scenario.

TrickSumo:
@ParidelPooya Thanks for the update.

Agree on behaviour:
```
(err as StepError).cause?.name === "StepInterruptedError"
```

With the required changes in the new PR, we can now close this PR.

PR #579, opened 2026-05-26
fix(sdk): remove runtime bundled detection from UserAgent header

ayushiahjolia:
*Issue #, if available:* N/A

*Description of changes:*
Customers (including Powertools) are experiencing bundling failures after v1.1.3 due to the runtime detection logic in `version.ts`. The code uses `node:url`, `node:path`, `__filename`, and `new Function("return import.meta")` to detect whether the SDK is running from the Lambda bundled runtime directory.

This breaks across CJS, ESM, and ESBuild bundling configurations.

### Changes

Remove the `isInLambdaRuntime()` function and all associated imports from `version.ts`, reverting to the clean state after #522. The file now exports two simple constants with no runtime detection:

- `SDK_NAME = "aws-durable-execution-sdk-js"`
- `SDK_VERSION = process.env.NPM_PACKAGE_VERSION || "0.0.0"`

### What this does NOT revert

PR #522 (UserAgent format fix removing `@` character) is preserved - that was a valid fix for AWS SDK formatting rules.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/utils/constants/version.ts
+++ b/packages/aws-durable-execution-sdk-js/src/utils/constants/version.ts
@@ -7,67 +7,11 @@
  *
  * SDK_NAME is a fixed string matching the cross-SDK convention:
  * aws-durable-execution-sdk-\{language\}
- * Alternate version if SDK is bundled into the Lambda runtime.
  *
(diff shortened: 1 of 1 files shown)


nvasiu (approved):

ayushiahjolia:
Instead of removing this, we will work on a fix.

PR #581, opened 2026-05-26
fix(sdk): remove esm-shim banner from ESM dist

ParidelPooya:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### Issue Link, if available

n/a — discovered while bumping `@aws/durable-execution-sdk-js` from 1.1.2 → 1.1.5 in a downstream consumer.

### Description

This PR has grown in scope since the first push to address review feedback. Three commits, each focused:

#### 1. Remove the `@rollup/plugin-esm-shim` banner from the ESM dist

`@aws/durable-execution-sdk-js@1.1.3 / 1.1.4 / 1.1.5` ship a `dist/index.mjs` with a top-level `cjsUrl.fileURLToPath(import.meta.url)` shim banner injected by `@rollup/plugin-esm-shim`. When a downstream consumer re-bundles the .mjs into CJS via esbuild (the default for AWS CDK `NodejsFunction` with `OutputFormat.CJS`), the literal `import.meta.url` survives into the CJS output, where `import.meta` is `undefined`, and Lambda module init crashes before the handler runs:

```
TypeError: The "path" argument must be of type string or an instance of URL. Received undefined
    at Object.fileURLToPath (node:internal/url:1606:11)
    at Object.<anonymous> (/var/task/index.js:1:931)
```

The shim was emitted because `version.ts` references `__filename` as a bare identifier inside `isInLambdaRuntime()`. Once any source file does so, `@rollup/plugin-esm-shim` injects an unconditional banner at the top of every emitted `.mjs`. Removing the plugin from the rollup config is sufficient to fix the bundling crash — the source can keep using bare `__filename` (it's a real CJS global at runtime in the dist-cjs build).

#### 2. Make ESM Lambda runtime detection actually work

Review on the first push pointed out that dropping the `__filename` fast-path would break CJS detection. While restoring the fast-path I noticed the ESM branch reads `import.meta` via `new Function("return import.meta")()`, which runs its body in non-module scope where `import.meta` is unavailable. The inner call throws, the outer try/catch swallows it, and ESM consumers silently get the un-suffixed SDK_VERSION.

Instrumented the SDK source with `console.log` statements inside `isInLambdaRuntime()` and deployed a test stack with two durable Lambdas (one CJS, one ESM bundled by CDK `NodejsFunction`). CloudWatch confirmed:

- **Before** (Function-constructor approach): ESM Lambda logged `Function-constructor threw: Cannot use 'import.meta' outside a module`, `currentFilePath` stayed `undefined`.
- **After** (this PR's fix, using top-level `import.meta.url`): ESM Lambda logged `branch=esm:import.meta.url, moduleFilePath=/var/task/index.mjs`. CJS Lambda continues logging `branch=cjs:__filename, moduleFilePath=/var/task/index.js`.

Both branches now read the SDK module's actual file path, so the `dirname(...).startsWith(LAMBDA_RUNTIME_DIR)` comparison runs against real input.

##### Instrumentation diffs and CloudWatch output

Two `console.log`-based probes were run in sequence: one against the unfixed source to diagnose, one against the fixed source to verify. Neither is committed.

**Probe 1** — instrumentation applied to the original `version.ts` (Function-constructor branch):

```diff
 function isInLambdaRuntime(): boolean {
   try {
     // ...NODE_ENV=test early-return omitted...
     let currentFilePath: string | undefined;
+    let branch: string;

     if (typeof __filename !== "undefined") {
       currentFilePath = __filename;
+      branch = "cjs:__filename";
     } else {
+      branch = "esm:Function-constructor";
       try {
         const getImportMeta = new Function("return import.meta");
         const importMeta = getImportMeta();
         if (importMeta && importMeta.url) {
           currentFilePath = fileURLToPath(importMeta.url);
         }
-      } catch {}
+      } catch (e) {
+        console.log(
+          "[isInLambdaRuntime] Function-constructor threw:",
+          (e as Error).message,
+        );
+      }
     }

+    console.log(
+      "[isInLambdaRuntime] branch=" + branch,
+      JSON.stringify({ currentFilePath, runtimeDir }),
+    );
     // ...startsWith comparison unchanged...
```

CloudWatch (Probe 1, against unfixed SDK):

```
[CJS Lambda]
[isInLambdaRuntime] branch=cjs:__filename
{"currentFilePath":"/var/task/index.js","runtimeDir":"/var/runtime"}

[ESM Lambda]
[isInLambdaRuntime] Function-constructor threw: Cannot use 'import.meta' outside a module
[isInLambdaRuntime] branch=esm:Function-constructor
{"runtimeDir":"/var/runtime"}
```

The ESM line confirms the Function-constructor diagnosis: `currentFilePath` never gets set, so detection silently returns `false`.

**Probe 2** — instrumentation applied to the fixed `version.ts` (top-level `import.meta.url`):

```diff
 let moduleFilePath: string | undefined;
+let _branch: string;
 if (typeof __filename !== "undefined") {
   moduleFilePath = __filename;
+  _branch = "cjs:__filename";
 } else if (typeof import.meta?.url === "string") {
   moduleFilePath = fileURLToPath(import.meta.url);
+  _branch = "esm:import.meta.url";
+} else {
+  _branch = "none";
 }
+console.log(
+  "[version] branch=" + _branch,
+  JSON.stringify({ moduleFilePath, runtimeDir }),
+);
```

CloudWatch (Probe 2, against the fix from this PR):

```
[CJS Lambda]
[version] branch=cjs:__filename
{"moduleFilePath":"/var/task/index.js","runtimeDir":"/var/runtime"}

[ESM Lambda]
[version] branch=esm:import.meta.url
{"moduleFilePath":"/var/task/index.mjs","runtimeDir":"/var/runtime"}
```

`moduleFilePath` now populates in both module systems. The `startsWith(/var/runtime)` comparison still returns `false` for both Lambdas because the test stack puts the bundled SDK at `/var/task/`, not in a Lambda Runtime layer at `/var/runtime/` — that's the expected outcome for a non-bundled SDK install. The probe's job is to confirm the upstream input (`moduleFilePath`) is correctly populated, not the downstream classification.

The previous `version.ts` had a `process.env.NODE_ENV === "test"` short-circuit so it could be loaded by Jest without exercising `__filename`. With the new top-level `import.meta.url` reference, ts-jest (which forces `module: commonjs`) refuses to compile the file with `TS1343: 'import.meta' is only allowed when 'module' is es2020+`. Rather than work around the compiler, this PR adds a Jest manual mock at `src/utils/constants/__mocks__/version.ts` that returns hardcoded `SDK_NAME` / `SDK_VERSION` values during tests, and wires it through `moduleNameMapper`. That means:

- Jest never compiles the real `version.ts`, so the ts-jest TS1343 error doesn't fire.
- The `NODE_ENV=test` short-circuit is no longer needed (and is removed); Jest can never reach the real detection code at all.
- `__mocks__/` is exempted from the package's kebab-case filename rule (Jest requires the literal directory name).

#### 3. Add integration tests covering all three regressions

- **`packages/esm-integration-test`** — mirrors the existing `cjs-integration-test`. Direct ESM `import()` of the published `dist/index.mjs`, no bundler involved.
- **`packages/cdk-bundling-integration-test`** — reproduces the AWS CDK `NodejsFunction` bundling path end-to-end. Five scenarios (CJS, ESM, ESM + require polyfill, CJS + bundleAwsSDK, ESM + bundleAwsSDK), each bundling a Lambda-shaped handler with `aws-cdk-lib/NodejsFunction`'s default esbuild flags and require/import-loading the result. Confirmed to fail against the unpatched `v1.x` SDK with the production crash and pass after.
- **`packages/lambda-runtime-detection-integration-test`** — spawns child Node processes (one CJS, one ESM) with controlled `LAMBDA_RUNTIME_DIR` and asserts the SDK_VERSION baked into `LambdaClient.config.customUserAgent` carries (or doesn't carry) the `-bundled` suffix. Four scenarios; needed because the manual Jest mock means the real detection logic isn't exercised by unit tests.

### Demo/Screenshots

`head -20 packages/aws-durable-execution-sdk-js/dist/index.mjs` before and after:

Before (1.1.5):
```js
import { fileURLToPath } from 'node:url';
import { dirname } from 'node:path';


// -- Shims --
import cjsUrl from 'node:url';
import cjsPath from 'node:path';
import cjsModule from 'node:module';
const __filename = cjsUrl.fileURLToPath(import.meta.url);
const __dirname = cjsPath.dirname(__filename);
const require = cjsModule.createRequire(import.meta.url);
```

After:
```js
import { OperationType, OperationStatus, OperationAction, ... } from '@aws-sdk/client-lambda';
import { EventEmitter } from 'events';
import { AsyncLocalStorage } from 'async_hooks';
import { createHash } from 'crypto';
import { Console } from 'node:console';
import util from 'node:util';
import { fileURLToPath } from 'node:url';
import { dirname } from 'node:path';
```

The `// -- Shims --` block and its top-level `import.meta.url` references are gone.

### Checklist

- [x] I have filled out every section of the PR template
- [x] I have thoroughly tested this change

#### Unit Tests

Existing SDK unit tests (793/793) pass.

#### Integration Tests

Three new packages added under `packages/`, all wired into the root `npm test` script:

- `esm-integration-test` (plain ESM `import` of the dist).
- `cdk-bundling-integration-test` (5 esbuild scenarios mirroring AWS CDK `NodejsFunction`'s defaults).
- `lambda-runtime-detection-integration-test` (4 spawn-based scenarios covering CJS+ESM × runtime-dir-matches/doesn't-match).

#### Real Lambda

Deployed a CDK stack with two durable Lambdas (`OutputFormat.CJS` and `OutputFormat.ESM`) that import the SDK and run `withDurableExecution`. Both invoked successfully via their `:live` aliases on `nodejs:22.DurableFunction.v16` runtime, returning the durable step result with `status: "success"` in CloudWatch — confirming the bundling fix and the runtime-detection branches both work in a real Lambda environment. Stack has been torn down.

#### AWS Powertools e2e

The original reproducer for this regression was the AWS Powertools idempotency e2e suite, which started failing on the bump from `^1.1.3` → `^1.1.5`. Built a tarball from this branch, vendored it into Powertools, and re-ran the e2e workflow — green: https://github.com/aws-powertools/powertools-lambda-typescript/actions/runs/26477242071

#### Examples

Not applicable — this is a build-pipeline / runtime-detection fix.

[#546]: https://github.com/aws/aws-durable-execution-sdk-js/pull/546
[#562]: https://github.com/aws/aws-durable-execution-sdk-js/pull/562

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,6 @@
         "@eslint/compat": "^1.3.0",
         "@eslint/js": "^9.29.0",
         "@microsoft/api-documenter": "^7.28.1",
-        "@rollup/plugin-esm-shim": "^0.1.8",
         "@rollup/plugin-json": "^6.1.0",
         "@rollup/plugin-typescript": "^12.1.4",
         "@tsconfig/node20": "^20.1.5",
@@ -45,8 +44,6 @@
     },
     "node_modules/@aws-crypto/crc32": {
       "version": "5.2.0",
-      "resolved": "https://registry.npmjs.org/@aws-crypto/crc32/-/crc32-5.2.0.tgz",
-      "integrity": "sha512-nLbCWqQNgUiwwtFsen1AdzAtvuLRsQS8rYgMuxCrdKf9kOssamGLuPwyTY9wyYblNr9+1XM8v6zoDTPPSIeANg==",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-crypto/util": "^5.2.0",
@@ -59,8 +56,6 @@
     },
     "node_modules/@aws-crypto/sha256-browser": {
       "version": "5.2.0",
-      "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-browser/-/sha256-browser-5.2.0.tgz",
-      "integrity": "sha512-AXfN/lGotSQwu6HNcEsIASo7kWXZ5HYWvfOmSNKDsEqC4OashTp8alTmaz+F7TC2L083SFv5RdB+qU3Vs1kZqw==",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-crypto/sha256-js": "^5.2.0",
@@ -74,8 +69,6 @@
     },
     "node_modules/@aws-crypto/sha256-browser/node_modules/@smithy/is-array-buffer": {
       "version": "2.2.0",
-      "resolved": "https://registry.npmjs.org/@smithy/is-array-buffer/-/is-array-buffer-2.2.0.tgz",
-      "integrity": "sha512-GGP3O9QFD24uGeAXYUjwSTXARoqpZykHadOmA8G5vfJPK0/DC67qa//0qvqrJzL1xc8WQWX7/yc7fwudjPHPhA==",
       "license": "Apache-2.0",
       "dependencies": {
         "tslib": "^2.6.2"
@@ -86,8 +79,6 @@
     },
     "node_modules/@aws-crypto/sha256-browser/node_modules/@smithy/util-buffer-from": {
       "version": "2.2.0",
-      "resolved": "https://registry.npmjs.org/@smithy/util-buffer-from/-/util-buffer-from-2.2.0.tgz",
-      "integrity": "sha512-IJdWBbTcMQ6DA0gdNhh/BwrLkDR+ADW5Kr1aZmd4k3DIF6ezMV4R2NIAmT08wQJ3yUK82thHWmC/TnK/wpMMIA==",
       "license": "Apache-2.0",
       "dependencies": {
         "@smithy/is-array-buffer": "^2.2.0",
@@ -99,8 +90,6 @@
     },
     "node_modules/@aws-crypto/sha256-browser/node_modules/@smithy/util-utf8": {
       "version": "2.3.0",
-      "resolved": "https://registry.npmjs.org/@smithy/util-utf8/-/util-utf8-2.3.0.tgz",
-      "integrity": "sha512-R8Rdn8Hy72KKcebgLiv8jQcQkXoLMOGGv5uI1/k0l+snqkOzQ1R0ChUBCxWMlBsFMekWjq0wRudIweFs7sKT5A==",
       "license": "Apache-2.0",
       "dependencies": {
         "@smithy/util-buffer-from": "^2.2.0",
@@ -112,8 +101,6 @@
     },
     "node_modules/@aws-crypto/sha256-js": {
       "version": "5.2.0",
-      "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-js/-/sha256-js-5.2.0.tgz",
-      "integrity": "sha512-FFQQyu7edu4ufvIZ+OadFpHHOt+eSTBaYaki44c+akjg7qZg9oOQeLlk77F6tSYqjDAFClrHJk9tMf0HdVyOvA==",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-crypto/util": "^5.2.0",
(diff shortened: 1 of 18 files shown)


ayushiahjolia:
Fix is merged.
v1.x - https://github.com/aws/aws-durable-execution-sdk-js/pull/585
main - https://github.com/aws/aws-durable-execution-sdk-js/pull/587

PR #593, opened 2026-05-29
fix(ci): resolve flaky integration test failures in deploy script

ayushiahjolia:
Fixes three sources of flaky integration test failures in the deploy script.

#### Changes
1. Wait for newly created functions to become active
After CreateFunctionCommand, regular (non-capacity-provider) functions go through a Pending state before becoming invocable. The deploy script previously returned immediately after creation, causing downstream tests to fail with "Function not found" when the test job started before functions transitioned to Active.

- Added a wait-for-active poll (up to 60 retries) after creating non-capacity-provider functions, matching the existing pattern used for capacity-provider functions.

2. Fix capacity provider vCPU limit detection
When the capacity provider hits its maxVCpuCount limit, the function enters Failed state. The existing code only checked LastUpdateStatusReasonCode.CapacityProviderScalingLimitExceeded, but the actual error sometimes arrives with a different/missing reason code - only the message string contains "maxVCpuCount limit". This caused the deploy to crash instead of taking the delete-and-retry path.

- Added a fallback check on the error message string in both detection points.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/scripts/deploy-lambda.ts
+++ b/packages/aws-durable-execution-sdk-js-examples/scripts/deploy-lambda.ts
@@ -523,6 +523,44 @@ async function main(): Promise<void> {
             useCapacityProvider,
             selectedRuntime,
           );
+
+          // Wait for newly created function to become active
+          if (!useCapacityProvider) {
+            console.log("Waiting for function to enter active state");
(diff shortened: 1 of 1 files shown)


ayushiahjolia:
this might be fixed by https://github.com/aws/aws-durable-execution-sdk-js/pull/698

PR #596, opened 2026-06-02
feat(sdk): Wire plugin into execution lifecycle

SilanHe:
*Issue #, if available:* #545

*Description of changes:*

Wire the DurableInstrumentationPlugin through the core execution path:

- Add plugin param to DurableExecution interface and CheckpointManager
- Wire plugin through withDurableExecution → runHandler → CheckpointManager
- Add onInvocationStart/End, wrapInvocation, and onOperationChange hook calls at the invocation level
- Add enrichLogContext support to DefaultLogger for plugin-provided log context fields
- Type-narrow wrapInvocation to return Promise<DurableExecutionInvocationOutput>
- Export plugin types from the public API barrel

Other changes:
- runAsCallback now swallows plugin errors and ultimately runs customerFn if every plugin throws. If there is one customerFn error is always propagated so that plugins can't be used to alter the output of invocation.
- InvocationInfo has been split, now only ExecutionEndInfo depends on InvocationBaseInfo which doesn't have the isFirstInvocation field.

Handler-level wiring (step, invoke, wait, callback, runInChildContext) will follow in a subsequent PR.


packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
```diff
@@ -84,125 +103,170 @@ async function runHandler<
...
-    });
-
-    // Wait for all pending checkpoints to complete
-    try {
-      await durableExecution.checkpointManager.waitForQueueCompletion();
-      log("✅", "All pending checkpoints completed");
-    } catch (error) {
-      log("⚠️", "Error waiting for checkpoint completion:", error);
-    }
+  const executeInvocation =
```
SilanHe, line 111: it looks like a complicated diff but I've just nested the whole thing into this new async function.

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
```diff
@@ -61,13 +62,22 @@ export interface AttemptEndInfo extends AttemptInfo {
...
   executionArn: string;
+}
+
+/**
+ * Information about a durable execution invocation, including whether it is
+ * the first invocation (execution mode) or a subsequent replay.
+ *
+ * @experimental This interface is experimental and may be changed or removed in future releases.
+ */
+export interface InvocationInfo extends InvocationBaseInfo {
```
ayushiahjolia, line 96: what about pending invocation status?

packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts
```diff
@@ -74,6 +75,8 @@ export class CheckpointManager implements Checkpoint {
     private stepDataEmitter: EventEmitter,
     private logger: DurableLogger,
     private finishedAncestors: Set<string>,
+    private plugin: DurableInstrumentationPlugin = {},
+    private requestId: string = "",
```
ayushiahjolia: can we make it required when plugin present?

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
SilanHe: ah I don't have this field in the InvocationInfo. Successful and Failed invocations according to your description in [java sdk](https://github.com/aws/aws-durable-execution-sdk-java/blob/main/sdk/src/main/java/software/amazon/lambda/durable/plugin/InvocationStatus.java) are encoded through the onExecutionEnd hook. For example, see [here](https://github.com/aws/aws-durable-execution-sdk-js/blob/4ef03a923b7236b6c4d9f41459cf4909ee4e2904/packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts#L318).

Retries and pending are not currently differentiated.

[Where onInvocationEnd is called](https://github.com/aws/aws-durable-execution-sdk-js/blob/4ef03a923b7236b6c4d9f41459cf4909ee4e2904/packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts#L332)

I see that you and Frank aligned on the InvocationStatus and discarded the onExecutionEnd function.

packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts
SilanHe: ah DurableInstrumentationPlugin is actually always present anyway so I can remove the default values here to make them both required.

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
SilanHe: The Typescript SDK doesn't have the RETRYING status on the invocation: https://github.com/aws/aws-durable-execution-sdk-js/blob/4ef03a923b7236b6c4d9f41459cf4909ee4e2904/packages/aws-durable-execution-sdk-js/src/types/core.ts#L89

zhongkechen: RETRYING is something specific in the plugin interface that indicates an uncaught exception is thrown from the function and the backend will retry

SilanHe: yeah I'm adding it in my next commit for this PR

packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
SilanHe: I've also added the onInvocationStart and oninvocationEnd calls where applicable

packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
```diff
@@ -51,10 +58,23 @@ async function runHandler<
...
+    createDefaultLogger(
+      executionContext,
+      plugin.enrichLogContext?.bind(plugin),
+    ),
     new Set<string>(),
+    plugin,
+    executionContext.requestId,
   );
 
+  const invocationInfo: InvocationInfo = {
```
zhongkechen, line 70: It's named `InvocationStartInfo` in Python. Actually all the types for `on*Start*` methods have `Start` in their names. And in `InvocationStartInfo` there is one additional field `start_time` in Python.

packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
```diff
@@ -84,125 +108,181 @@ async function runHandler<
...
+        if (resultType === "termination") {
+          log("🛑", "Returning termination response");
+
+          plugin.onInvocationEnd?.({
+            ...invocationInfo,
+            status: PluginInvocationStatus.PENDING,
+            executionInput: customerHandlerEvent,
+            executionResult: undefined,
+            executionError: undefined,
+            operations: executionContext._stepData,
```
zhongkechen, line 215: Wow, this is a lot of internal information. Are we sure we want to expose this internal data structure to plugin?

zhongkechen (approved):

packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
SilanHe: ahh interesting, I'm not opposed to it. I took a quick glance at Java and it looks like we don't have the "Start" in the names. Let's align on this with Ayushi and do the change in another PR.

packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
SilanHe: not sure yet, but I can see how useful these are. These are part of the fields Pooya planned to expose from the spec.

https://github.com/aws/aws-durable-execution-sdk-js/blame/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/documents/INSTRUMENTATION_PLUGIN.md#L109

PR #599, opened 2026-06-03
feat(ci): add automatic runtime release workflow

nvasiu:
*Issue #, if available:*

N/A

*Description of changes:*

Added a new GitHub action which automatically runs whenever there is a new release for the sdk from the `v1.x` branch. The action will create a second companion release, pointing to the same commit as the original release.

The companion releases will be tagged with a consistent format, and will be used by an automated testing harness to run integration tests on new sdk versions being released to the Lambda runtime.

### DIFF
--- a/.github/workflows/runtime-release.yml
+++ b/.github/workflows/runtime-release.yml
@@ -1,0 +1,36 @@
+name: Create companion runtime release
+
+on:
+  release:
+    types: [published]
+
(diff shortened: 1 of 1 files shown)


nvasiu:
No need for the automated companion release at this time, our runtime testing currently uses the original `sdk-x.y.z` release.

PR #606, opened 2026-06-04
feat(sdk): Wire plugin hooks into run-in-child-context handler

SilanHe:
*Issue #, if available:*

*Description of changes:*

Add plugin instrumentation to the run-in-child-context handler with hooks for operation lifecycle (onOperationFirstStart, onOperationStart, wrapChildContextFn, onOperationFirstEnd).

- Pass plugin from DurableContext into createRunInChildContextHandler
- Add plugin test coverage for run-in-child-context handler hooks

Key decisions: For virtual child context, onOperationStart and onOperationEnd will be called on every replay since we do not checkpoint the results of the virtual child context.


SilanHe:
closing in favour of https://github.com/aws/aws-durable-execution-sdk-js/pull/616

packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts
```diff
@@ -303,6 +318,13 @@ export const executeChildContext = async <T, Logger extends DurableLogger>(
       Type: OperationType.CONTEXT,
       Name: name,
     });
+    await plugin.onOperationStart?.({
+      ...opInfo,
+      status: OperationStatus.STARTED,
+      isReplay: false,
+    });
+  } else {
+    await plugin.onOperationStart?.({ ...opInfo, isReplay: true });
```
zhongkechen, line 327: call plugin for virtual context?

ayushiahjolia (approved):

packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts
SilanHe: I'm open to discussions about this but I think it makes sense to trigger the onOperationStart and onOperationEnd hooks. The virtual child context, though not a metered operation, is still an "operation".

ayushiahjolia: I think we should only trigger those during first run and not replay, similar behavior to normal operations.

PR #610, opened 2026-06-05
feat(ci): derive npm dist-tag from version, keep latest on highest line

ParidelPooya:
Replaces the GitHub Release prerelease flag with a per-package, version- derived dist-tag decision so v1 and v2 lines can be maintained in parallel:

- 'latest' only ever moves to a version GREATER than the current 'latest' (read from the registry). A stable backport below 'latest' is published to a release-line tag v<major> and never moves 'latest'.
- A prerelease is published to 'beta' only if it is greater than 'latest'; an old-line prerelease is rejected before any publish.
- An already-published version is skipped (idempotent re-runs).
- A genuine registry read failure aborts the release (never guesses).
- Post-publish verification confirms the targeted dist-tag points at the published version.

Decision logic is a pure, unit-tested module (decide-dist-tag.js) with a vendored, spec-correct semver comparison (handles 1.10.0 > 1.9.0 and prerelease precedence). Adds a registry-aware mock + e2e harness covering all paths. Removes the obsolete flag-based tag suite (test-publish-tags.yml, assert-publish-tags.js, mock-npm.js and their unit tests) and drops the now- unused prerelease arg from npm-publish.yml.


.github/workflows/iterate-publish-npm.sh
```diff
@@ -1,23 +1,129 @@
...
+  [ -d "$dir" ] || continue
+  name=$(pkg_field "$dir" name)
+  version=$(pkg_field "$dir" version)
+
+  if ! packument=$(fetch_packument "$name"); then
+    echo "ERROR: aborting release; could not determine current 'latest' for $name (no packages published)."
+    exit 1
+  fi
+
+  decision=$(node "$DECIDE" "$version" "$packument")
```
yaythomas: Non-blocking: this captures stdout but doesn't check node's exit code. If `decide-dist-tag.js` throws (e.g., on an invalid version string in `package.json`), `decision` is empty, the `case` below falls through to the `*)` branch, and we end up publishing with `--tag ""` (which would fail with a confusing error, possibly after earlier packages in the loop already published).

With `set -uo pipefail` (no `-e`), an empty assignment doesn't abort the script. Suggest something like:

```bash
if ! decision=$(node "$DECIDE" "$version" "$packument"); then
  echo "ERROR: decide-dist-tag.js failed for $name@$version" >&2
  exit 1
fi
```

yaythomas (dismissed):

.github/workflows/iterate-publish-npm.sh
ParidelPooya: Addressed in revision2

yaythomas (approved):

PR #616, opened 2026-06-09
feat(sdk): checkpoint plugin hooks

SilanHe:
*Issue #, if available:*

*Description of changes:* 

Following the approach we're taking in Python and Java where we trigger the plugin hooks based on data we received from CheckpointDurableExecution and the invoke lambda call from the backend.

Will follow up with testing-sdk change


packages/aws-durable-execution-sdk-js/src/context/execution-context/execution-context.ts
```diff
@@ -79,6 +97,27 @@ export const initializeExecutionContext = async (
 
   log("📝", "Loaded step data:", stepData);
 
+  // Dispatch inter-invocation hooks for operations that updated between invocations
```
SilanHe, line 100: crux of change for operations completing between invocations

packages/aws-durable-execution-sdk-js/src/types/core.ts
```diff
@@ -69,6 +69,20 @@ export interface DurableExecutionInvocationInput {
      */
     NextMarker?: string | undefined;
   };
+
+  /**
+   * Optional array of operation IDs that were updated between invocations.
+   *
+   * When present, the SDK dispatches inter-invocation plugin hooks (`onOperationStart`,
```
ayushiahjolia, line 76: There won't be any case where we emit onOperationStart for operation in updatedOperationIds list, right?

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
```diff
@@ -20,16 +20,17 @@ export enum PluginInvocationStatus {
...
  * Information about a durable operation.
  *
+ * The `Id` and `ParentId` fields always contain hashed values as returned
+ * by the checkpoint response.
+ *
  * @experimental This interface is experimental and may be changed or removed in future releases.
  */
 export interface OperationInfo {
   Id: string;
-  HashedId: string;
```
SilanHe: I'm also removing this HashedId, since we plan to only provide hashedId

packages/aws-durable-execution-sdk-js/src/types/core.ts
SilanHe: I think it's technically possible if the Lambda invocation crashes prior to getting the response to a CheckpointDurableExecution call and at the same time the backend crashes prior to seeing that event.

ayushiahjolia (dismissed):

packages/aws-durable-execution-sdk-js/src/utils/operation/operation.ts
```diff
@@ -42,50 +34,23 @@ export function toAttemptInfo(
...
-      defaults.HashedParentId ??
-      (info.ParentId ? hashId(info.ParentId) : undefined);
-  if (!info.StartTimestamp) info.StartTimestamp = defaults.StartTimestamp;
-  if (!info.EndTimestamp) info.EndTimestamp = defaults.EndTimestamp;
-  return info;
+export function extractErrorFromOperation(
+  operation: Operation,
+): Error | undefined {
+  if (operation.Status === OperationStatus.FAILED) {
+    const errorData =
```
ayushiahjolia, line 78: We don't need to emit errorData for TimedOut, Stopped?

packages/aws-durable-execution-sdk-js/src/utils/operation/operation.ts
```diff
@@ -1,28 +1,20 @@
...
 /**
  * Converts an Operation to an OperationInfo.
+ * The Id and ParentId fields always contain hashed values as returned
+ * by the checkpoint response.
  *
  * @experimental This function is experimental and may be changed or removed in future releases.
  */
 export function toOperationInfo(operation?: Operation): OperationInfo {
   return {
     Id: operation?.Id ?? "",
```
zhongkechen, line 18: Is this unhashed raw id? How would people/plugins match this ID with the operation IDs from history?

packages/aws-durable-execution-sdk-js/src/context/execution-context/execution-context.ts
```diff
@@ -79,6 +97,27 @@ export const initializeExecutionContext = async (
...
+    for (const operationId of event.updatedOperationIds) {
+      const operation = stepData[operationId];
+      if (!operation) continue;
+
+      const status = operation.Status;
+      if (status && TERMINAL_STATUSES.includes(status as OperationStatus)) {
+        const error = extractErrorFromOperation(operation);
+        plugin.onOperationEnd?.({ ...toOperationInfo(operation), error });
+      } else if (status === OperationStatus.STARTED) {
+        plugin.onOperationStart?.(toOperationInfo(operation));
```
zhongkechen, line 115: Backend starts a new operation or update an existing operation status to STARTED? Is this possible?

packages/aws-durable-execution-sdk-js/src/context/execution-context/execution-context.ts
```diff
@@ -79,6 +97,27 @@ export const initializeExecutionContext = async (
 
   log("📝", "Loaded step data:", stepData);
 
+  // Dispatch inter-invocation hooks for operations that updated between invocations
+  if (
+    event.updatedOperationIds &&
+    event.updatedOperationIds.length > 0 &&
+    plugin
+  ) {
+    for (const operationId of event.updatedOperationIds) {
```
zhongkechen, line 106: Is it OK to process the list at the beginning of invocation? I think the other SDKs will check this list when the operation is actually replayed

packages/aws-durable-execution-sdk-js/src/utils/operation/operation.ts
SilanHe: this is hashedId now

packages/aws-durable-execution-sdk-js/src/context/execution-context/execution-context.ts
SilanHe: I think it's technically possible if the Lambda invocation crashes prior to getting the response to a CheckpointDurableExecution call and at the same time the backend crashes prior to seeing that event.

packages/aws-durable-execution-sdk-js/src/context/execution-context/execution-context.ts
SilanHe: Since we're not providing an order of operations implicitly via the raw Id and raw parentId, I don't think this is a problem.

packages/aws-durable-execution-sdk-js/src/utils/operation/operation.ts
SilanHe: oops, yeah I do need to add these

SilanHe:
dropping this in favour of per handler approach

PR #618, opened 2026-06-10
feat(sdk): add Workflow Insight plugin scaffold with contracts and types

ParidelPooya:
Adds the aws-durable-execution-sdk-js-insight package with:

- Plugin contracts doc (InsightExporter interface, WorkflowInsightConfig, WorkflowInsightRecord schema, content filtering, operation overrides)
- POC plan and observability pillars positioning docs
- Minimal TypeScript implementation: workflowInsight() factory returning a DurableInstrumentationPlugin with no-op onInvocationStart, onInvocationEnd, and onOperationChange listeners
- package.json, tsconfig.json, src/types.ts, src/index.ts

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight/src/index.ts
+++ b/packages/aws-durable-execution-sdk-js-insight/src/index.ts
@@ -1,0 +1,64 @@
+import type { Operation } from "@aws-sdk/client-lambda";
+import type { WorkflowInsightConfig } from "./types";
+
+export type {
+  InsightExporter,
+  WorkflowInsightConfig,
+  WorkflowInsightRecord,
+  OperationRecord,
+  ContentConfig,
+  OperationOverride,
+} from "./types";
+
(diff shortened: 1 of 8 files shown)


packages/aws-durable-execution-sdk-js-insight/src/index.ts
```diff
@@ -0,0 +1,64 @@
+import type { Operation } from "@aws-sdk/client-lambda";
```
yaythomas, line 1: only the sdk declare this dependency in package.json

probably doesn't matter too much, since this will always be used with the sdk?

yaythomas (approved):

packages/aws-durable-execution-sdk-js-insight/src/index.ts
ParidelPooya: Good point, will address it.

Issue #595, opened 2026-06-01
[Bug]: Child operation should call ser & des before return the result


wangyb-A:
### Expected Behavior

I would like the the child operatoin to return the deseralized of [seralized result](https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts#L336) on first run. So that the first run and replay will have same result.

### Actual Behavior

The first run will return the raw result. The seralized result was checkpointed. So replay will return the deseralized result.

### Steps to Reproduce

```
import {
  DurableContext,
  withDurableExecution,
  Serdes,
  SerdesContext,
} from "@aws/durable-execution-sdk-js";

const uppercaseSerdes: Serdes<string> = {
  serialize: async (value: string | undefined, _context: SerdesContext) => {
    if (value === undefined) return undefined;
    return value.toUpperCase();
  },
  deserialize: async (data: string | undefined, _context: SerdesContext) => {
    return data;
  },
};

export const handler = withDurableExecution(
  async (event: any, context: DurableContext) => {
    const result = await context.runInChildContext(
      "serdes-child",
      async (childContext: DurableContext) => {
        return await childContext.step(async () => event as string);
      },
      { serdes: uppercaseSerdes },
    );
    return result;
  },
);
```

### SDK Version

1.1.7

### Node.js Version

22.x

### Is this a regression?

No

### Additional Context

Please note that this might be a breaking change.

wangyb-A:
Please don't forget to check the vitual context serdes for child operation

PR #626, opened 2026-06-11
fix(sdk): ensure child context result matches replay on first run

ayushiahjolia:
*Issue #, if available:*
https://github.com/aws/aws-durable-execution-sdk-js/issues/595

*Description of changes:*
- Child context results now go through the same serialize/deserialize round-trip on first run that replay would apply, ensuring deterministic values regardless of when or how many times the execution resumes.

**run-in-child-context-handler.ts**
- Small payloads: return `deserialize(serialize(result))` to match replay
- Large payloads (ReplayChildren): return raw result since replay
  re-executes the child function
- Virtual contexts: return raw result since they never checkpoint

**concurrent-execution-handler.ts**
- Default map/parallel to `createBatchResultSerdes` so errors survive the round-trip

**batch-result.ts**
- Added `serializeBatchError`/`reconstructBatchError` to preserve error type and cause chain through serialization

*Testing:*
Add tests using an asymmetric serdes that exercise the standard and virtual-context paths; existing tests only used lossless serdes and could not detect the divergence.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts
+++ b/packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts
@@ -253,15 +253,20 @@ export const handleCompletedChildContext = async <
       undefined,
       entityId, // parentId
     );
-    return await runWithContext(entityId, entityId, () =>
+
+    const replayedResult = await runWithContext(entityId, entityId, () =>
       fn(durableChildContext),
     );
+
+    // Large payloads: replay re-executes the child function, so return raw result.
+    return replayedResult;
   }
 
   log("⏭️", "Child context already finished, returning cached result:", {
     entityId,
   });
 
+  // Small payloads: replay deserializes the checkpoint, so match that here.
   return await safeDeserialize(
     serdes,
     result,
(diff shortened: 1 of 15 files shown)


wangyb-A (dismissed):

packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts
```diff
@@ -432,7 +437,21 @@ export const executeChildContext = async <T, Logger extends DurableLogger>(
       await plugin.onOperationEnd?.({ ...opInfo, isReplay: true });
     }
 
-    return result;
+    // Large payloads: replay re-executes the child function, so return raw result.
+    // Virtual contexts: never checkpoint, so no replay path to match.
+    if (replayChildren || isVirtual) {
+      return result;
```
wangyb-A, line 443: Do we need to add ser / des here?

packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts
```diff
@@ -253,15 +253,20 @@ export const handleCompletedChildContext = async <
...
       entityId, // parentId
     );
-    return await runWithContext(entityId, entityId, () =>
+
+    const replayedResult = await runWithContext(entityId, entityId, () =>
       fn(durableChildContext),
     );
+
+    // Large payloads: replay re-executes the child function, so return raw result.
+    return replayedResult;
```
wangyb-A, line 262: Do we need to add ser / des here?

wangyb-A (approved):

packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts
wangyb-A: will be done in separate PR

packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts
wangyb-A: will be done in separate PR

Issue #620, opened 2026-06-11
[Feature]: OnInvocationEnd hook should finish before Lambda execution ends


wangyb-A:
### What would you like?

Currently we are not await `onInvocationEnd()`, if lambda execution finish early, the hook call will be terminated as well.

We should make sure the onInvocationEnd hook is done before the lambda finish. While we do not cause durable execution time increase because of this.

onOpStart and onOpEnd are also fire and forget, add a document says it's plugin's responsibility to decide to wait / not wait for the hook end or not.

### Possible Implementation

1. Checkpoint execution result before we call OnINvocationEnd
2. Await onInvocationEnd

### Is this a breaking change?

No

### Does this require an RFC?

No

PR #633, opened 2026-06-12
feat(sdk): await all non-wrap plugin lifecycle hooks

wangyb-A:
*Issue #, if available:* #620

Makes the non-`wrap` plugin lifecycle hooks awaitable and has the SDK await
  them before proceeding past each lifecycle point. Previously only
  `onInvocationEnd` was awaited (and it accepted `void | Promise<void>`); the
  `onOperation*` hooks were fire-and-forget. This generalizes the contract so a
  plugin's async work is guaranteed to complete before the SDK moves on — e.g.
  before the next operation runs, or before the Lambda response is returned and
  the execution environment is frozen. This is required for the OpenTelemetry
  plugin, whose exporters flush telemetry asynchronously and must finish before
  the environment freezes.

  ## Changes
  - **`types/plugin.ts`**: `onInvocationStart`, `onInvocationEnd`, `onOperationStart`,
    `onOperationEnd`, `onOperationAttemptStart`, `onOperationAttemptEnd`, and
    `onOperationChange` now return `Promise<void>`. (`wrapInvocation`,
    `wrapChildContextFn`, `wrapOperationAttemptFn`, and `enrichLogContext` are
    unchanged.) Added TSDoc describing the always-await contract and how to opt
    into fire-and-forget.
  - **`plugin-runner.ts`**: collapsed `run` (fire-and-forget) and `runAwaitable`
    into a single always-awaiting helper. Hooks still run concurrently across
    plugins via `Promise.allSettled`, and plugin errors (sync or async) are
    swallowed so a misbehaving plugin never affects execution.
  - **Call sites**: `await` the hooks in `with-durable-execution.ts`,
    `step-handler.ts`, and `checkpoint-manager.ts`.
  - **`workflow-insight`**: hooks made `async` to match the new signatures.

  ## Plugin author contract
  - All non-`wrap` hooks are awaited before the SDK proceeds; time spent in a
    hook adds to the invocation's wall-clock duration.
  - To opt out (fire-and-forget), start the work and return without awaiting it.
    Such work is **best-effort**: in Lambda the environment freezes once the
    invocation returns, so unfinished detached work may not complete.
  - Do **not** fire-and-forget in `onInvocationEnd` (it runs right before the
    response). Use it instead to flush/await any background work started by
    earlier hooks.

  By submitting this pull request, I confirm that you can use, modify, copy, and
  redistribute this contribution, under the terms of your choice.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/handlers/step-handler/step-handler.ts
+++ b/packages/aws-durable-execution-sdk-js/src/handlers/step-handler/step-handler.ts
@@ -268,7 +268,7 @@ export const createStepHandler = <Logger extends DurableLogger>(
             stepData = context.getStepData(stepId);
             const operationInfo = toOperationInfo(stepData);
             backfillOperationInfo(operationInfo, opInfo);
-            plugin.onOperationStart?.({
+            await plugin.onOperationStart?.({
               ...operationInfo,
               isReplay: false,
             });
@@ -282,11 +282,11 @@ export const createStepHandler = <Logger extends DurableLogger>(
                 Type: OperationType.STEP,
                 Name: name,
               })
-              .then(() => {
+              .then(async () => {
                 stepData = context.getStepData(stepId);
                 const operationInfo = toOperationInfo(stepData);
                 backfillOperationInfo(operationInfo, opInfo);
-                plugin.onOperationStart?.({
+                await plugin.onOperationStart?.({
                   ...operationInfo,
                   isReplay: false,
                 });
@@ -295,7 +295,7 @@ export const createStepHandler = <Logger extends DurableLogger>(
         } else {
           const operationInfo = toOperationInfo(stepData);
           backfillOperationInfo(operationInfo, opInfo);
-          plugin.onOperationStart?.({ ...operationInfo, isReplay: true });
+          await plugin.onOperationStart?.({ ...operationInfo, isReplay: true });
         }
 
         try {
@@ -319,7 +319,7 @@ export const createStepHandler = <Logger extends DurableLogger>(
           );
           const attemptInfo = toAttemptInfo(stepData, currentAttempt);
           backfillOperationInfo(attemptInfo, opInfo);
-          plugin.onOperationAttemptStart?.(attemptInfo);
+          await plugin.onOperationAttemptStart?.(attemptInfo);
           let result: T;
           const stepFn = (): Promise<T> => fn(stepContext);
           result = (await runWithContext(
@@ -360,8 +360,8 @@ export const createStepHandler = <Logger extends DurableLogger>(
             },
           );
           backfillOperationInfo(attemptEndInfo, opInfo);
-          plugin.onOperationAttemptEnd?.(attemptEndInfo);
-          plugin.onOperationEnd?.({ ...attemptEndInfo, isReplay: false });
+          await plugin.onOperationAttemptEnd?.(attemptEndInfo);
+          await plugin.onOperationEnd?.({ ...attemptEndInfo, isReplay: false });
           checkpoint.markOperationState(
             stepId,
             OperationLifecycleState.COMPLETED,
@@ -421,8 +421,8 @@ export const createStepHandler = <Logger extends DurableLogger>(
               },
             );
             backfillOperationInfo(attemptEndInfo, opInfo);
-            plugin.onOperationAttemptEnd?.(attemptEndInfo);
-            plugin.onOperationEnd?.({
+            await plugin.onOperationAttemptEnd?.(attemptEndInfo);
+            await plugin.onOperationEnd?.({
               ...attemptEndInfo,
               isReplay: false,
               error: error instanceof Error ? error : new Error(String(error)),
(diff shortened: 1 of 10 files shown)


packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
```diff
@@ -78,7 +78,6 @@ type FileSystemEnvelope =
   | { data: string }
   | { file: string; preview?: Record<string, unknown> };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
ParidelPooya, line 81: nit: is this chage related to the plugin onInvocationEnd change?

packages/aws-durable-execution-sdk-js/src/utils/serdes/preview.ts
```diff
@@ -109,7 +109,7 @@ function isMatched(path: string, fields: PreviewField[] | undefined): boolean {
  *
  * @public
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+
```
ParidelPooya: The removal left a blank line between the @public TSDoc block and export function buildPreview.

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
SilanHe: This line gets removed every time we run "npm run build" locally, perhaps there's a setting we can flip so that it doesn't happen every time

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
```diff
@@ -133,7 +133,24 @@ export interface DurableInstrumentationPlugin {
...
+   * @remarks
+   * Because the SDK awaits this hook, any time spent here directly increases
+   * the wall-clock duration of the current Lambda invocation.
+   * Errors thrown (or rejected promises) are swallowed by the SDK
+   * and never affect the execution outcome.
+   *
+   * @param info - Details about the completed invocation, including status,
+   * result or error, input, and operations.
+   */
+  onInvocationEnd?(info: InvocationEndInfo): void | Promise<void>;
```
SilanHe: I've also raised this offline but I'm confused in general by this change because the onInvocationEnd hook returns void which implicitly states that it should not be an async function. 

I believe it's fine to have the hooks to be sync functions because it will guarantee the order in which they are called. When a customer adds a plugin, they are expecting there to be overhead whenever the hooks are called. There's no need to introduce this Promise<void> return type in onInvocationEnd since none of the other hooks return Promise<void>, so they are also implicitly supposed to be sync functions. If the customer decides to implement an async function within the function signature, they may choose to do so and still return void. In order to guarantee that all the function hooks are called prior to an invocation end, they will need manually track these Promises in their plugin.

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
```diff
@@ -125,22 +125,80 @@ export interface OperationChangeInfo extends InvocationBaseInfo {
...
+ * wall-clock duration of the current Lambda invocation.
+ *
+ * **Opting out (fire-and-forget):** to avoid blocking the SDK, start the
+ * background work inside the hook but return without awaiting it, so the
+ * returned promise resolves immediately:
+ *
+ * ```typescript
+ * onOperationEnd(info) {
+ *   // Start background work without awaiting it, then return.
+ *   void this.exporter.export(info).catch(() => {});
```
SilanHe, line 150: is this example comprehensive? what is `this.exporter.export()` ?

packages/aws-durable-execution-sdk-js/src/handlers/step-handler/step-handler.ts
```diff
@@ -282,11 +282,11 @@ export const createStepHandler = <Logger extends DurableLogger>(
                 Type: OperationType.STEP,
                 Name: name,
               })
-              .then(() => {
+              .then(async () => {
```
SilanHe, line 285: this is awkward because this doesn't necessarily mean onOperationStart will be awaited in this case. It's probably fine in this case since the checkpoint itself is more of a fire-and-forget as well

SilanHe (approved):

packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
ParidelPooya: I need to check it too.

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
SilanHe: we discussed offline. Due to how Javscript await/async works, it's better to change all the return types to Promise<void>. We'll be awaiting on all plugin hooks, so it's still kind of similar to what I'm talking about in the comment. If a customer wants to fire and forget, they can call an async function which contains the work they would like to do that is not awaited within their function.

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
ParidelPooya: It's only an example. we don't have this.exporter.export in our code

packages/aws-durable-execution-sdk-js/src/handlers/step-handler/step-handler.ts
ParidelPooya: Let's fix it.

Issue #622, opened 2026-06-11
[Bug]: PromiseCombinatorError is not reconstructed by DurableOperationError.fromErrorObject


ParidelPooya:
### Expected Behavior

An error serialized with ErrorType: "PromiseCombinatorError" should be reconstructed as a PromiseCombinatorError on replay, so instanceof checks behave identically on first execution and on replay.

### Actual Behavior

errors/durable-error/durable-error.ts adds case "CallbackTimeoutError" and case "CallbackSubmitterError" to fromErrorObject, but has no case "PromiseCombinatorError". The default branch returns StepError. If a PromiseCombinatorError propagates to an enclosing operation that fails
  and checkpoints, it round-trips as StepError on replay — non-deterministic error identity.

### Steps to Reproduce

  1. Inside a handler, run a failing combinator inside an outer child context/step that re-throws, so the outer operation checkpoints an error with ErrorType: "PromiseCombinatorError".
  2. Force a replay (resume the execution).
  3. Catch the error and check the type.

  Minimal unit-level repro:

  const obj = { ErrorType: "PromiseCombinatorError", ErrorMessage: "all failed" };
  const e = DurableOperationError.fromErrorObject(obj as any);
  // Expected: e instanceof PromiseCombinatorError === true
  // Actual:   e instanceof StepError === true

### SDK Version

2.0.0-alpha.1

### Node.js Version

22.x

### Is this a regression?

No

PR #639, opened 2026-06-16
fix(sdk): add PromiseCombinatorError case to fromErrorObject

ParidelPooya:
PromiseCombinatorError was missing from the switch in DurableOperationError.fromErrorObject, causing it to deserialize as StepError on replay. This broke instanceof checks and introduced non-deterministic error identity.

Changes:
- Add case "PromiseCombinatorError" to fromErrorObject switch
- Export PromiseCombinatorError from package index
- Add integration tests for round-trip serialization

*Issue #, if available:*
#622 

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/errors/durable-error/durable-error.ts
+++ b/packages/aws-durable-execution-sdk-js/src/errors/durable-error/durable-error.ts
@@ -75,6 +75,12 @@ export abstract class DurableOperationError extends Error {
           cause,
           errorObject.ErrorData,
         );
+      case "PromiseCombinatorError":
+        return new PromiseCombinatorError(
+          errorObject.ErrorMessage || "Promise combinator failed",
+          cause,
+          errorObject.ErrorData,
+        );
       default:
         return new StepError(
           errorObject.ErrorMessage || "Unknown error",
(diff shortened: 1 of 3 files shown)


packages/aws-durable-execution-sdk-js/src/errors/durable-error/durable-error.ts
```diff
@@ -75,6 +75,12 @@ export abstract class DurableOperationError extends Error {
           cause,
           errorObject.ErrorData,
         );
+      case "PromiseCombinatorError":
+        return new PromiseCombinatorError(
+          errorObject.ErrorMessage || "Promise combinator failed",
+          cause,
+          errorObject.ErrorData,
+        );
       default:
```
yaythomas, line 84: nit: default silently coerces into StepError. maybe add a `satisfies never`?

yaythomas (approved):

packages/aws-durable-execution-sdk-js/src/errors/durable-error/durable-error.ts
ParidelPooya: The better safeguard is what we already have: the integration test. If someone adds a new error subclass with a new errorType but forgets the switch case, the round-trip test pattern will catch it (and we can add a parametrized test that iterates over all subclasses to guarantee
  coverage).

Issue #578, opened 2026-05-26
[Bug]: replay mode is always set to ExecutionMode after virtual context


zhongkechen:
### Expected Behavior

The replay mode of a context correctly reflect the status if the current operation has been replayed or not.

### Actual Behavior

The replay mode always switches to ExecutionMode If there is a virtual context operation

```
1. Step (Completed) <- replay mode is set to ExecutionMode
2. Virtual Context (Completed) 
3. Print logs <- always output the logs
4. Step (Completed)
```

Example code:

```js
export const handler = withDurableExecution(
  async (event: any, context: DurableContext) => {
    const result = await context.runInChildContext(
      async (childContext: DurableContext) => {
        const stepResult = await childContext.step(async () => {
          return "virtual child step completed";
        });
        return stepResult;
      },
      { virtualContext: true }, // Skip checkpointing to save operation costs
    );
    context.logger.info("After virtual context");  // <-- always output the message
    await context.wait({ seconds: 2 });
    return result;
  },
);
```

The code to set the replay mode is here:
https://github.com/aws/aws-durable-execution-sdk-js/blob/09ce060b1cb3a7943a9a354228e1dd750942f068/packages/aws-durable-execution-sdk-js/src/context/durable-context/durable-context.ts#L231-L238

### Steps to Reproduce

1. Create durable function with 
    1. run a virtual context
    2. output logs
    3. wait
2. The step after virtual context always logs even if in replay mode

### SDK Version

latest

### Node.js Version

22.x

### Is this a regression?

No

SilanHe:
I think creating a virtual runInChildContext in the root user function could be considered a no-op/anti-pattern. It's the reason why we provide abstractions that help the customer such as map and parallel.

https://docs.aws.amazon.com/durable-execution/sdk-reference/operations/child-context/

In the example, creating a "virtual" runInChildContext is effectively a no op. Personally, I don't think we need to fix this as it's more of a "feature". I do think that we'll likely need to document this clearly in the behaviour of our new plugin hooks(#545).

zhongkechen:
There should be an easy fix of this issue without asking users to avoid this feature. I think we can change the way of creating operation IDs for virtual contexts, maybe like `{prefix}-v1` , `{prefix}-v2 `, ... so that the non-virtual operation IDs like `{prefix}-1`, `{prefix}-2`, ... are always checkpointed. This will require us to create a separate operation counter for virtual operations

SilanHe:
[...]

PR #644, opened 2026-06-17
fix(sdk): replay mode when preceded by virtual context

SilanHe:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-js/issues/578

*Description of changes:* The replay mode always switched to ExecutionMode If current operation is followed by a virtual context operation meaning that we were often emitting more logs than necessary on replays. In this PR, we fix the ExecutionMode to check whether the next "genuine/non virtual" operation has been checkpointed.

This effectively deduplicates the logging where the current following logical operation is a virtual child context operation and the operations after the virtual child context are already completed. It doesn't change the behaviour of the logging for operations run in parallel.


SilanHe:
spoke with Pooya offline,

The proposed approach above is still missing the following case:

```
1 step
log("hello")
1.1 virtual context {
   1.1_1 wait
}
1.2 virtual context
```

if "1.1_1 wait" is present in context.getStepData("1.1_1"), log("hello") will still be emitted.

Instead, he proposes the following change. Preserve the current stepCounter for both real and virtual operations. In `checkAndUpdateReplayMode`, if the next operation doesn't exist in the stepData, then we try to check context.getStepData("nextStepId_1"). If that step is not there, then we switch to ExecutionMode.

```
1 step
log("hello")
2 virtual context {
   2_1 wait
}
3 step
```

This design knowingly ignores the cases where we have 2 nested virtual context as it's an edge case with no customer use:

```
1 step
log("hello")
2 virtual context {
   2_1 virtual context {

   }
} 
3 step
```

SilanHe:
it's not a pressing bug fix so closing the PR for now

PR #646, opened 2026-06-17
fix(sdk): preserve errorData across nested runInChildContext boundaries

ParidelPooya:
DurableOperationError.toErrorObject only copied this.errorData, so when an error carrying errorData crossed two or more runInChildContext boundaries the payload was lost. Each boundary wraps the error in a fresh ChildContextError whose own errorData is undefined, overwriting the deeper value on serialization.

toErrorObject now walks the cause chain (bounded to 10 hops) when its own errorData is undefined and surfaces the first DurableOperationError.errorData it finds. No public API or type changes.

Adds a run-in-child-context/error-data-propagation example and test that throws a CallbackError with errorData through two nested child contexts and asserts the payload is still recoverable at the catch site.

*Issue #, if available:*
#524

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/src/examples/run-in-child-context/error-data-propagation/run-in-child-context-error-data-propagation.ts
+++ b/packages/aws-durable-execution-sdk-js-examples/src/examples/run-in-child-context/error-data-propagation/run-in-child-context-error-data-propagation.ts
@@ -1,6 +1,5 @@
 import {
   CallbackError,
-  ChildContextError,
   DurableContext,
   DurableOperationError,
   withDurableExecution,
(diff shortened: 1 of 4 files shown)


yaythomas (commented):
tests failing

packages/aws-durable-execution-sdk-js-examples/src/examples/run-in-child-context/error-data-propagation/run-in-child-context-error-data-propagation.ts
```diff
@@ -0,0 +1,51 @@
...
+    try {
+      await context.runInChildContext(
+        "outer-child",
+        async (outerChild: DurableContext) => {
+          await outerChild.runInChildContext(
+            "inner-child",
+            async (innerChild: DurableContext) => {
+              await innerChild.step("throw-with-error-data", async () => {
+                throw new CallbackError("cb failed", undefined, SENTINEL);
+              });
```
yaythomas: if you add this it will de-flake the test, maybe?

```
  { retryStrategy: () => ({ shouldRetry: false }) },
```

yaythomas (changes requested):

packages/aws-durable-execution-sdk-js-examples/src/examples/run-in-child-context/error-data-propagation/run-in-child-context-error-data-propagation.ts
ParidelPooya: Fixed

yaythomas (approved):

PR #648, opened 2026-06-17
fix(sdk): apply serdes round-trip across all runInChildContext modes

ParidelPooya:
Previously runInChildContext only ran the deserialize(serialize(result)) round-trip for small (checkpointed) payloads. Large payloads (ReplayChildren mode) and virtual contexts returned the raw in-memory result instead. With a non-identity serdes this meant the value a caller received depended on payload size or the virtual flag, an inconsistent and surprising contract.

All three modes now return deserialize(serialize(result)) so callers observe the same serdes round-trip regardless of payload size or virtual context. The ReplayChildren replay path re-serializes/deserializes the re-executed result so first-run and replay stay consistent. Virtual contexts still never checkpoint; they just apply the round-trip in memory.

Adds run-in-child-context serdes examples and integ tests for all three modes (small, large payload, virtual) that use an asymmetric uppercasing serdes to prove the round-trip is applied. Updates the SDK unit and integration tests that previously asserted the raw-return behavior. No public API changes.


yaythomas (changes requested):
functional code is :shipit: 

tests failing, though. 

> Cannot find module './run-in-child-context-serdes.js'.

The new examples import the runtime value uppercaseSerdes from ../serdes/run-in-child-context-serdes, which is itself a catalog example. Which rollup isn't vendoring.

Might want to put uppercaseSerdes into a shared non-handler module (e.g. run-in-child-context/shared-serdes.ts)

yaythomas (approved):

Issue #625, opened 2026-06-11
[Bug]: FileSystem-serdes does not sanitize entityId used as a filename


ParidelPooya:
### Expected Behavior

The on-disk filename derived from entityId should be safe for all valid operation names, just as the directory derived from the execution ARN is encoded via encodeURIComponent. 

**Right now it's not an issue but it could cause issue for DAG project as we will use operationName to generate entityId.**

### Actual Behavior

writeToFile uses encodeURIComponent(context.durableExecutionArn) for the directory but the raw entityId for the filename (${context.entityId}.json). User-provided operation names flow into entityId; a name containing / creates/escapes directories or fails with ENOENT, and other
  special characters are unhandled.

### Steps to Reproduce

  context.configureSerdes({ defaultSerdes: createFileSystemSerdes("/mnt/s3") });
  // name contains a slash
  await dagContext.step("invoices/2026", async () => bigPayload);
  // Expected: value persisted to a single safe file
  // Actual:   write targets "/mnt/s3/<arn>/invoices/2026.json" → ENOENT (parent "invoices" not created) / path escape

### SDK Version

2.0.0-alpha.1

### Node.js Version

22.x

### Is this a regression?

No

PR #651, opened 2026-06-18
feat(sdk): sanitize filesystem serdes paths and add pathEncoding option

ParidelPooya:
The filesystem serdes derived the on-disk directory from `encodeURIComponent(durableExecutionArn)` but used the raw `entityId` for the file name (`${entityId}.json`). An entity ID containing characters such as `/` or `..` could therefore escape or fail to create the intended directory. This is latent today (entity IDs are counter-based step IDs) but becomes a real problem once user-controlled operation names flow into the entity ID. 

Changes:
- Always sanitize the entity ID before using it as a file name.
- Add a `pathEncoding` config option (new public `FileSystemPathEncoding` enum) to choose how the ARN and entity ID are turned into path segments:
  - `URI` (default): the per-execution directory is a compact, human-navigable path parsed from the ARN — `<functionName>/<executionName>/<invocationId>` (these are stable across an execution's invocations and already filesystem-safe), and the file name is `encodeURIComponent(entityId)`. Falls back to encoding the whole ARN when it is not a durable-execution ARN.
  - `HASH`: the whole ARN (directory) and the entity ID (file name) are each replaced by their SHA-256 hex digest — fixed length and always safe, regardless of the characters or length of the original value.
- Export `FileSystemPathEncoding` from the package index.
- Add unit tests covering the compact directory layout, the non-standard ARN fallback, entity-ID sanitization, and the HASH layout.

*Issue #, if available:*
#625

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
+++ b/packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
@@ -1,5 +1,6 @@
 import { mkdir, readFile, writeFile } from "node:fs/promises";
 import { join } from "node:path";
+import { createHash } from "node:crypto";
 import { SerdesContext, AnySerdes } from "./serdes";
 import { CHECKPOINT_SIZE_LIMIT_BYTES } from "../constants/constants";
 export {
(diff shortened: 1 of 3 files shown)


packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts
```diff
@@ -78,15 +116,108 @@ type FileSystemEnvelope =
...
+  arn: string,
+  pathEncoding: FileSystemPathEncoding,
+): string {
+  if (pathEncoding === FileSystemPathEncoding.URI) {
+    const parts = parseDurableExecutionArn(arn);
+    if (parts) {
+      return join(
+        basePath,
+        parts.functionName,
+        parts.executionName,
```
yaythomas, line 196: restricted to Pattern: `[a-zA-Z0-9-_]+` by api, so no encode necessary

yaythomas (approved):

PR #652, opened 2026-06-18
feat(sdk): align linear retry strategy with exponential helper

ParidelPooya:
Bring createLinearRetryStrategy to parity with createRetryStrategy by switching it to a config-object API and adding the capabilities it was missing. Both helpers now share a common module so their behavior cannot drift apart.

Changes:
- Add src/utils/retry/retry-common with shared applyJitter, resolveRetryableErrors, isErrorRetryable, and finalizeDelaySeconds (jitter + integer/min-1s clamp).
- Refactor createRetryStrategy to delegate to the shared module (no behavior change).
- Rewrite createLinearRetryStrategy with a LinearRetryStrategyConfig object using Duration values, and add maxDelay capping, jitter (default FULL), and retryableErrors/retryableErrorTypes filtering.
- Update retryPresets.linear to the new API with jitter NONE to keep its documented deterministic 1s-5s delays.
- Export LinearRetryStrategyConfig from the package entry point.
- Expand linear retry tests: maxDelay, non-second Duration units, all jitter modes, integer/min-1 clamping, and error filtering.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/index.ts
+++ b/packages/aws-durable-execution-sdk-js/src/index.ts
@@ -79,7 +79,10 @@ export {
   createRetryStrategy,
   RetryStrategyConfig,
 } from "./utils/retry/retry-config";
-export { createLinearRetryStrategy } from "./utils/retry/linear-retry-strategy/linear-retry-strategy";
+export {
+  createLinearRetryStrategy,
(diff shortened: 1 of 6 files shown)


yaythomas (approved):
this is a breaking change

PR #653, opened 2026-06-19
feat(sdk): add executionInput to InvocationInfo (plugin interface)

ParidelPooya:
## Summary

Moves `executionInput` from `InvocationEndInfo` up to `InvocationInfo` so plugins receive the execution input on `onInvocationStart` — not only at invocation end.

## Motivation

Observability plugins (like Workflow Insight) need to include the execution input in their first emitted record. Currently `executionInput` is only available in `onInvocationEnd`, which means:
- The initial RUNNING record has no input
- If the execution suspends (wait/callback) before completing, no record ever has the input

### DIFF
--- a/packages/aws-durable-execution-sdk-js-otel/src/__tests__/plugin.test.ts
+++ b/packages/aws-durable-execution-sdk-js-otel/src/__tests__/plugin.test.ts
@@ -39,6 +39,7 @@ function makeInvocationInfo(
     requestId: TEST_REQUEST_ID,
     executionArn: TEST_ARN,
     isFirstInvocation: true,
+    executionInput: {},
     operations: {},
     ...overrides,
   };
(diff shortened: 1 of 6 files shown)


packages/aws-durable-execution-sdk-js-otel/src/__tests__/plugin.test.ts
```diff
@@ -51,11 +52,11 @@ function makeInvocationEndInfo(
     requestId: TEST_REQUEST_ID,
     executionArn: TEST_ARN,
     isFirstInvocation: true,
+    executionInput: {},
```
ayushiahjolia: don't we need to remove it from makeInvocationEndInfo here?

ParidelPooya: executionInput here is required because InvocationEndInfo extends InvocationInfo, and we moved executionInput to InvocationInfo (it's no longer in InvocationEndInfo's own definition). So this isn't a duplicate, it's satisfying the inherited required field. Added a comment to clarify

ayushiahjolia (approved):

PR #663, opened 2026-06-29
feat(sdk): add Workflow Insight plugin and VS Code Explorer extension

ParidelPooya:
Two new packages for workflow observability:

## @aws/durable-execution-sdk-js-insight

SDK plugin that captures durable execution state and exports curated WorkflowInsightRecord to customer-owned destinations. Features:

- 14 built-in exporters: LambdaLog, CloudWatch Logs, S3, DynamoDB, Aurora (MySQL+PostgreSQL), Redshift, OpenSearch, Firehose, EventBridge, SQS, Timestream, OTel, HTTP/webhook, File (EFS)
- Coalescing export scheduler (no overlapping calls, intermediate updates dropped since each record is a full snapshot)
- wrapInvocation-based drain (records guaranteed delivered before Lambda returns)
- emitMode: finished-only or in-progress
- recordType discriminator for easy filtering
- Handles non-Date timestamps from the real durable runtime

## aws-durable-execution-sdk-js-insight-vscode

VS Code extension for querying Workflow Insight data in plain English.

- Two-step wizard: fetch data → visualize
- 3 query providers: CloudWatch Logs Insights, DynamoDB (PartiQL), Aurora PostgreSQL (full SQL)
- 3 LLM backends: Amazon Bedrock, GitHub Copilot, local Qwen 3B
- React + Cloudscape Design System UI
- 9 chart presets via Vega-Lite + export (PNG/SVG)
- Self-correction loop (feeds query errors back to LLM)
- Settings modal with per-destination conditional fields
- LLM suggests chart types based on query result shape


yaythomas:
1) webview.js (88,337 lines) is a generated webview bundle committed to source; the package .gitignore excludes dist/ and node_modules/ but not media/webview.js, so the build artifact is being checked in.

2) aurora-exporter.ts (INSERT INTO ${this.table}) and the Redshift exporter. Values are correctly parameterized via RDS Data API named params (good), but this.table comes from config and is string-interpolated. It

ParidelPooya:
> 2\. aurora-exporter.ts (INSERT INTO ${this.table}) and the Redshift exporter. Values are correctly parameterized via RDS Data API named params (good), but this.table comes from config and is string-interpolated.

Addressed both

yaythomas (dismissed):

yaythomas (dismissed):

yaythomas (approved):

PR #664, opened 2026-06-29
feat(sdk): plugin operation status enum is updated to const enum

SilanHe:
*Issue #, if available:*

*Description of changes:* 

1. Adding PluginOperationStatus const enum which maps directly from OperationStatus of `@aws-sdk/client-lambda`.
2. All exported enum from plugin.ts are also const enum now.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/types/plugin.ts
+++ b/packages/aws-durable-execution-sdk-js/src/types/plugin.ts
@@ -9,11 +9,30 @@ import { DurableExecutionInvocationOutput } from "./core";
  *
  * @experimental This enum is experimental and may be changed or removed in future releases.
  */
-export enum PluginInvocationStatus {
+export const enum PluginInvocationStatus {
   SUCCEEDED = "SUCCEEDED",
   FAILED = "FAILED",
   PENDING = "PENDING",
   RETRYING = "RETRYING",
+}
+
+/**
+ * Status values for durable operations.
+ *
+ * These represent the lifecycle states that an operation can be in
+ * during durable execution. 
+ *
+ * @experimental This enum is experimental and may be changed or removed in future releases.
+ */
+export const enum PluginOperationStatus {
+  STARTED = "STARTED",
+  READY = "READY",
+  PENDING = "PENDING",
+  SUCCEEDED = "SUCCEEDED",
+  FAILED = "FAILED",
+  TIMED_OUT = "TIMED_OUT",
+  STOPPED = "STOPPED",
+  CANCELLED = "CANCELLED",
 }
 
 /**
@@ -27,7 +46,7 @@ export interface OperationInfo {
   type: string;
   subType?: string;
   parentId?: string;
-  status?: string;
+  status?: PluginOperationStatus;
   startTimestamp?: Date;
   endTimestamp?: Date;
   result?: string;
@@ -57,7 +76,7 @@ export interface AttemptInfo extends OperationInfo {
  *
  * @experimental This enum is experimental and may be changed or removed in future releases.
  */
-export enum AttemptEndInfoOutcome {
+export const enum AttemptEndInfoOutcome {
   SUCCEEDED = "SUCCEEDED",
   FAILED = "FAILED",
   RETRYING = "RETRYING",
--- a/package-lock.json
+++ b/package-lock.json
@@ -58,6 +58,74 @@
         "node": ">=16.0.0"
       }
     },
+    "node_modules/@aws-crypto/crc32c": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/crc32c/-/crc32c-5.2.0.tgz",
+      "integrity": "sha512-+iWb8qaHLYKrNvGRbiYRHSdKRWhto5XlZUEBwDjYNf+ly5SVYG6zEoYIdxvf5R3zyeP16w4PLBn3rH1xc74Rag==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@aws-crypto/util": "^5.2.0",
+        "@aws-sdk/types": "^3.222.0",
+        "tslib": "^2.6.2"
+      }
+    },
+    "node_modules/@aws-crypto/sha1-browser": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/sha1-browser/-/sha1-browser-5.2.0.tgz",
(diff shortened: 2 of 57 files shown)


packages/aws-durable-execution-sdk-js/src/types/plugin.ts
```diff
@@ -9,13 +9,32 @@ import { DurableExecutionInvocationOutput } from "./core";
...
+ */
+export const enum PluginOperationStatus {
+  STARTED = "STARTED",
+  READY = "READY",
+  PENDING = "PENDING",
+  SUCCEEDED = "SUCCEEDED",
+  FAILED = "FAILED",
+  TIMED_OUT = "TIMED_OUT",
+  STOPPED = "STOPPED",
+  CANCELLED = "CANCELLED",
```
SilanHe, line 35: Currently, I'm essentially mapping directly from `@aws-sdk/client-lambda` `OperationStatus` which is why there's other statuses which we're not directly assigning ourselves. We are getting some of these statuses from backend updates. If this is not ok, I'm happy to change.

Generally, we only really expect to see "STARTED" in onOperationStart calls. For onOperationEnd calls, we generally expect to see "SUCCEEDED" and "FAILED" statuses. ~~"READY" status is also possible in onOperationStart calls for subsequent retries for STEP type operation such as step and wait-for-condition.~~ The other terminal statuses such as TIMED_OUT, STOPPED and CANCELLED would only come from backend and are included just in case a customer calls StopDurableExecution or the execution times out.

ParidelPooya:
A TS enum compiles to a self-referencing IIFE:
```
  var PluginOperationStatus;
  (function (P) { P["STARTED"]="STARTED"; P["READY"]="READY"; /* ... */ })
    (PluginOperationStatus || (PluginOperationStatus = {}));
```
  Bundlers (Rollup/terser/esbuild) generally can't prove this is side-effect-free, so the whole object is retained even if a consumer references only one member — and it can't be
  split member-by-member. So your concern is correct: plain enum is the worst of the options for bundle size.

  const enum is the best for tree-shaking (fully inlined, zero runtime), but as covered it breaks JS consumers, Babel/esbuild/SWC, and isolatedModules users (including this repo's
  own otel/testing packages). So we want a third option that is both tree-shakeable and safe across the package boundary.

  Recommended fix: as const object + derived union type

  This is the smithy/AWS SDK v3 pattern. Replace the const enum declarations in types/plugin.ts with:
```
  /**
   * Status values for durable operations.
   *
   * These represent the lifecycle states an operation can be in during durable
   * execution. Mirrors `OperationStatus` from `@aws-sdk/client-lambda`.
   *
   * `STARTED` is what you'll see in onOperationStart; `READY` can also appear on
   * retries of STEP-type operations (step / wait-for-condition). onOperationEnd
   * normally reports `SUCCEEDED` or `FAILED`. The remaining terminal states
   * (`TIMED_OUT`, `STOPPED`, `CANCELLED`) originate from the backend, e.g. when a
   * customer calls StopDurableExecution or the execution times out.
   *
   * @experimental This API is experimental and may change or be removed.
   */
  export const PluginOperationStatus = {
    STARTED: "STARTED",
    READY: "READY",
    PENDING: "PENDING",
    SUCCEEDED: "SUCCEEDED",
    FAILED: "FAILED",
    TIMED_OUT: "TIMED_OUT",
    STOPPED: "STOPPED",
    CANCELLED: "CANCELLED",
  } as const;

  export type PluginOperationStatus =
    (typeof PluginOperationStatus)[keyof typeof PluginOperationStatus];
```
  Apply the same shape to PluginInvocationStatus and AttemptEndInfoOutcome.

  Why this satisfies every constraint:

  - Tree-shakeable. It's a plain const object literal, not an IIFE. Bundlers treat it as pure and can drop it entirely when unused. AWS SDK v3 deliberately switched from enum to
  this exact pattern for tree-shaking.
  - Safe across the package boundary. A real runtime object is emitted, so JS consumers, Babel/esbuild/SWC, and isolatedModules projects (your otel/testing packages) all work — no
  inlining required, no ambient-const-enum error.
  - Not a breaking change for the two already-public enums: PluginOperationStatus.STARTED, Object.values(...), etc. all still work at runtime.
  - No call-site changes. Because the value object and the type share the name, every existing PluginOperationStatus.STARTED and status?: PluginOperationStatus keeps compiling
  unchanged. The handler diffs in the PR stay as-is.
  - Maximum consistency with the @aws-sdk/client-lambda OperationStatus you're mirroring.

  The one ergonomic difference vs. a real enum: you can't use it as a type namespace for individual members (e.g. PluginOperationStatus.STARTED as a type), but the union type
  covers the normal need.

  Alternative (lightest possible): pure string-literal union

  If you don't need a runtime value object at all, a bare union is zero-runtime and perfectly tree-shaken:
```
  export type PluginOperationStatus =
    | "STARTED" | "READY" | "PENDING" | "SUCCEEDED"
    | "FAILED" | "TIMED_OUT" | "STOPPED" | "CANCELLED";
```
  Downside: producers/consumers write the raw strings (status: "STARTED"), and you'd have to change call sites that reference .STARTED. Given the PR already uses member access in
  the handlers, the as const object is the lower-churn choice.

  Recommendation

  Go with the as const object pattern for all three. It's the only option that fixes tree-shaking, keeps the runtime object, avoids breaking changes, and stays compatible with
  Babel/esbuild/isolatedModules consumers.

  Want me to apply this to the local clone (all three enums in plugin.ts) and run npm install && npm run build && npm test to confirm it compiles, tests pass, and the call sites
  are untouched? That would validate the approach before you suggest it on the PR.
  
  **the recommended as const fix is backward compatible for practically everyone (runtime identical, only member-as-type usage breaks), whereas the PR's const enum is a
  real backward-compatibility break for the two pre-existing enums. If you want, I can apply the as const version locally and build/test to confirm the call sites and existing
  behavior are untouched.**

ayushiahjolia (approved):

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
SilanHe: hmmm since const enum doesn't export an object anymore at runtime, I wonder if the cast from OperationStatus toPluginOperationStatus would still work. I don't think we cover the READY, TIMED_OUT, STOPPED or CANCELLED case in our tests.

Might be safer to just use regular enum or const objects.

svozza: +1 for a const object: if you use enums that means you can't use the the built in node feature that allows you to run TypeScript files directly: https://devblogs.microsoft.com/typescript/announcing-typescript-5-8-beta/#the---erasablesyntaxonly-option

PR #666, opened 2026-06-29
fix(sdk): use parameterized queries in RedshiftExporter

ParidelPooya:
Replace string interpolation with Redshift Data API named parameters (:name syntax + Parameters array). The previous approach used manual quote-escaping which is fragile and an injection surface since record_json contains arbitrary customer input/output.

SQL template is pre-built at construction time (immutable) — only parameters change per export call.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight/src/exporters/redshift-exporter.ts
+++ b/packages/aws-durable-execution-sdk-js-insight/src/exporters/redshift-exporter.ts
@@ -1,6 +1,7 @@
 import {
   RedshiftDataClient,
   ExecuteStatementCommand,
+  type SqlParameter,
 } from "@aws-sdk/client-redshift-data";
 import type { InsightExporter, WorkflowInsightRecord } from "../types";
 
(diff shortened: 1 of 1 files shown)


packages/aws-durable-execution-sdk-js-insight/src/exporters/redshift-exporter.ts
```diff
@@ -85,20 +86,44 @@ export class RedshiftExporter implements InsightExporter {
...
+    VALUES
+      (:execution_arn, :execution_name, :function_name, :status, :start_time, :end_time, :duration_ms, :record_json, :emitted_at)`;
   }
 
   async export(record: WorkflowInsightRecord): Promise<void> {
-    const fqTable = `${this.schema}.${this.table}`;
-    const sql = this.buildMerge(fqTable, record);
+    const parameters: SqlParameter[] = [
+      { name: "execution_arn", value: record.executionArn },
+      { name: "execution_name", value: record.executionName ?? "" },
```
yaythomas: https://docs.aws.amazon.com/redshift/latest/mgmt/data-api.html

> "Running SQL statements with parameters"):

> You can't set a value to NULL. The Data API interprets it as the literal string NULL.
> You can't set a zero length value. The Data API SQL statement fails. [example: {"name": "id", "value": ""}]

 the old esc(null)/num(null) emitted SQL NULL for these columns.

 the new code emits a failing zero-length parameter.

ParidelPooya: Addressed

yaythomas (approved):

PR #667, opened 2026-06-30
fix(sdk): enforce read-only queries in VS Code extension

ParidelPooya:
LLM-generated queries are executed against the customer's database with potentially write-capable credentials. A hallucinated or prompt-injected DELETE/DROP/UPDATE would execute unchecked.

Added assertReadOnly() that validates before execution:
- Must start with SELECT or WITH (CTEs)
- Rejects INSERT/UPDATE/DELETE/DROP/ALTER/CREATE/TRUNCATE/MERGE/etc.
- Rejects multiple statements (semicolon injection)

Applied to both Aurora (PostgreSQL) and DynamoDB (PartiQL) paths. CloudWatch Logs Insights is inherently read-only.

Defense-in-depth: customers should also configure read-only DB credentials for the extension.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight-vscode/src/queryValidator.ts
+++ b/packages/aws-durable-execution-sdk-js-insight-vscode/src/queryValidator.ts
@@ -1,0 +1,47 @@
+/**
+ * Validates that a SQL/PartiQL query is read-only (SELECT only).
+ * Rejects any statement that could mutate data.
+ *
+ * This is a defense-in-depth measure since queries are LLM-generated
+ * and could be hallucinated or prompt-injected. Customers should also
+ * configure their DB credentials with read-only permissions.
+ */
+
+const DANGEROUS_KEYWORDS =
+  /^\s*(INSERT|UPDATE|DELETE|DROP|ALTER|CREATE|TRUNCATE|MERGE|REPLACE|GRANT|REVOKE|EXEC|EXECUTE|CALL|COPY|UNLOAD)\b/i;
+const VALID_SELECT = /^\s*(SELECT|WITH)\b/i;
+
+/**
+ * Throws if the query is not a single read-only SELECT (or WITH...SELECT).
+ */
+export function assertReadOnly(query: string, engine: string): void {
+  const trimmed = query.trim();
+
+  if (!trimmed) {
+    throw new Error("Empty query.");
+  }
+
+  // Check for dangerous keywords at statement start
+  if (DANGEROUS_KEYWORDS.test(trimmed)) {
+    throw new Error(
+      `Refused to execute ${engine} query: only SELECT statements are allowed. ` +
+        `Got: "${trimmed.substring(0, 40)}..."`,
+    );
+  }
+
+  // Must start with SELECT or WITH (for CTEs)
+  if (!VALID_SELECT.test(trimmed)) {
+    throw new Error(
+      `Refused to execute ${engine} query: must start with SELECT or WITH. ` +
+        `Got: "${trimmed.substring(0, 40)}..."`,
+    );
+  }
+
+  // Reject multiple statements (semicolon followed by non-whitespace)
+  const withoutStrings = trimmed.replace(/'[^']*'/g, ""); // strip string literals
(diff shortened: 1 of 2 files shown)


packages/aws-durable-execution-sdk-js-insight-vscode/src/queryValidator.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Validates that a SQL/PartiQL query is read-only (SELECT only).
+ * Rejects any statement that could mutate data.
+ *
+ * This is a defense-in-depth measure since queries are LLM-generated
+ * and could be hallucinated or prompt-injected. Customers should also
+ * configure their DB credentials with read-only permissions.
+ */
+
+const DANGEROUS_KEYWORDS =
+  /^\s*(INSERT|UPDATE|DELETE|DROP|ALTER|CREATE|TRUNCATE|MERGE|REPLACE|GRANT|REVOKE|EXEC|EXECUTE|CALL|COPY|UNLOAD)\b/i;
```
yaythomas, line 11: anchored with ^\s*, so it only matches a mutating keyword as the first token. But VALID_SELECT intentionally allows WITH, and PostgreSQL CTEs can be data-modifying:

sql

> WITH t AS (DELETE FROM users RETURNING *) SELECT * FROM t;
> 

This passes all three checks and then runs through runAuroraQuery (ExecuteStatementCommand), which commits when not inside an explicit transaction. INSERT ... RETURNING and UPDATE ... RETURNING inside a CTE bypass it the same way.

Because a regex on the leading keyword can't see inside the CTE, I don't think string-matching can close this safely (scanning for the keyword anywhere would false-reject legitimate queries that mention delete in a column name or string literal). I'd suggest the real enforcement live at the DB layer, with the regex kept only as a cheap first filter:

Aurora/PostgreSQL: run inside a read-only transaction. 

BeginTransactionCommand → SET TRANSACTION READ ONLY (or BEGIN READ ONLY) → ExecuteStatementCommand → Commit/Rollback. 

PostgreSQL rejects any write at execution time with ERROR: cannot execute ... in a read-only transaction. That's authoritative rather than heuristic.
Continue to recommend (as you already do) read-only DB credentials as the primary control.

Note the DynamoDB/PartiQL path is fine here because PartiQL has no CTE/WITH support, so the WITH allowance is harmless there and INSERT|UPDATE|DELETE are caught at the start. The concern is specifically the Aurora path.

packages/aws-durable-execution-sdk-js-insight-vscode/src/queryValidator.ts
```diff
@@ -0,0 +1,47 @@
...
+  // Must start with SELECT or WITH (for CTEs)
+  if (!VALID_SELECT.test(trimmed)) {
+    throw new Error(
+      `Refused to execute ${engine} query: must start with SELECT or WITH. ` +
+        `Got: "${trimmed.substring(0, 40)}..."`,
+    );
+  }
+
+  // Reject multiple statements (semicolon followed by non-whitespace)
+  const withoutStrings = trimmed.replace(/'[^']*'/g, ""); // strip string literals
```
yaythomas, line 41: doesn't account for SQL escaped quotes ('') or PostgreSQL dollar-quoting ($$ ... $$). This won't create a write bypass (the multi-statement check only rejects), but it can mis-handle a body that contains ;, e.g. a dollar-quoted literal, and produce a confusing false rejection. If the read-only-transaction approach above is adopted, this whole semicolon heuristic becomes redundant and could be dropped.

packages/aws-durable-execution-sdk-js-insight-vscode/src/queryValidator.ts
ParidelPooya: The main line of defence is credentials with read-only permissions

packages/aws-durable-execution-sdk-js-insight-vscode/src/queryValidator.ts
ParidelPooya: I'll add more, but for now the main line of defence is credentials with read-only permissions.

yaythomas (approved):

PR #670, opened 2026-06-30
feat(sdk): add PluginOperationStatus as const object (tree-shakeable, backward-compatible)

ParidelPooya:
## Issue #, if available:
N/A

## Description of changes:

This is an alternative to #664 that delivers the same feature while keeping the public API **tree-shakeable _and_ backward compatible**.

### What it does
1. Adds a `PluginOperationStatus` type that mirrors `OperationStatus` from `@aws-sdk/client-lambda` (`STARTED`, `READY`, `PENDING`, `SUCCEEDED`, `FAILED`, `TIMED_OUT`, `STOPPED`, `CANCELLED`).
2. Types `OperationInfo.status` as `PluginOperationStatus` and wires the enum into `onOperationStart` / `onOperationEnd` in the step, wait-for-condition, and run-in-child-context handlers.
3. Converts `PluginInvocationStatus` and `AttemptEndInfoOutcome` to the same pattern for consistency.

### Why `as const` objects instead of `const enum`
`const enum` is fully erased at compile time and emits **no runtime object**, which:
- breaks plain-JS consumers and `Object.values(...)` (the symbol is `undefined` at runtime),
- breaks transpile-only / `isolatedModules` toolchains (Babel, esbuild, SWC) — including this repo's own `-otel` and `-testing` packages,
- is a **breaking change** for the already-published `PluginInvocationStatus` and `AttemptEndInfoOutcome` enums.

Using an `as const` object + derived union type (the exact pattern AWS SDK v3 uses for `OperationStatus`) keeps the values tree-shakeable (a plain object literal, not the un-shakeable `enum` IIFE) while still emitting a real runtime export, so existing consumers keep working.

### Bonus
Because `OperationStatus` is assignable to the resulting string union, the `as PluginOperationStatus` cast in `toOperationInfo` is no longer needed and was removed (`operation.ts` now matches `main`).

### Verification
- `npm run build` (rollup esm + cjs + tsc types) — confirmed `exports.PluginOperationStatus` is present in the runtime CJS bundle.
- `tsc --noEmit` for the SDK package and the `-otel` / `-testing` consumer packages (which use `isolatedModules`).
- `jest`: 1012 tests / 86 suites passing.
- `eslint`: no new issues.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts
+++ b/packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts
@@ -28,7 +28,10 @@ import {
 import { runWithContext } from "../../utils/context-tracker/context-tracker";
 import { DurablePromise } from "../../types/durable-promise";
 import { DurableLogger } from "../../types/durable-logger";
-import { DurableInstrumentationPlugin } from "../../types/plugin";
+import {
+  DurableInstrumentationPlugin,
+  PluginOperationStatus,
+} from "../../types/plugin";
 import {
   backfillOperationInfo,
   toOperationInfo,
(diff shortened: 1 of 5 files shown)


SilanHe (approved):
thanks!

ayushiahjolia (approved):

PR #671, opened 2026-06-30
ci(otel): Otel test exporter reset refactor

SilanHe:
*Issue #, if available:*

*Description of changes:* Added new assertion on the exact number of spans. Refactor tests to reset the exporter at the test level instead of within the handler so that we can assert on the number of spans exported for the execution of a durable lambda function provided in otel examples.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/src/examples/otel/child-context/otel-child-context.test.ts
+++ b/packages/aws-durable-execution-sdk-js-examples/src/examples/otel/child-context/otel-child-context.test.ts
@@ -1,10 +1,14 @@
-import { handler } from "./otel-child-context";
+import { handler, resetExporter } from "./otel-child-context";
 import { createTests } from "../../../utils/test-helper";
 import { SerializedSpan } from "../shared/otel-test-setup";
 
 createTests({
   handler,
   tests: (runner, { assertEventSignatures }) => {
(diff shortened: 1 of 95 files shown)


packages/aws-durable-execution-sdk-js-examples/src/examples/otel/child-context/otel-child-context.test.ts
```diff
@@ -16,6 +20,9 @@ createTests({
       expect(result.result).toBe("inner-1-result:inner-2-result");
 
       const { spans } = result;
+      // All spans: child-ctx (CONTEXT), inner-step-1 (op + attempt),
+      // inner-step-2 (op + attempt) = 5 spans
+      expect(spans.length).toBe(5);
```
zhongkechen, line 25: Why don't we assert what the spans actually are?

zhongkechen (approved):

packages/aws-durable-execution-sdk-js-examples/src/examples/otel/child-context/otel-child-context.test.ts
SilanHe: yeah I could do that haha, I just didn't. Since you approved, I'm happy to do it as follow up

PR #677, opened 2026-07-02
feat(sdk): per-store operationsByName index for Workflow Insight

ParidelPooya:
Lets users query executions by a *named operation's* metrics. `operations` stays a
canonical **array** in array-native stores; point-access stores (CloudWatch,
DynamoDB) carry a name-keyed **`operationsByName`** map instead; and the flexible
HTTP/OTel exporters make the choice configurable. Each store carries exactly one
operations representation.

Operation names repeat (loops/retries/`map` — the agentic-loop reuses
`context.step("invoke-model", …)`), so a name-keyed map can't be the canonical shape
without losing data, and dynamic keys break schema stores (OpenSearch mapping
explosion, unstable Athena schema). So the array is canonical; the map is a
per-exporter projection. Full rationale + multi-store query examples in
`docs/operations-shape.md`.

### What changed

  **Plugin core**
  - New `OperationSummary` / `OperationsFormat` types and
    `buildOperationsByName()` / `withOperationsByName()` / `applyOperationsFormat()`.
  - `buildOperationsByName` Aggregates in one pass over the operations, then materializes the smaller distinct-name map - no grouping or sorting:
    - Metrics aggregate across all occurrences: `count`, `min`/`max`/`totalDurationMs`,
      `failedCount`, `maxAttempt`.
    - `type`/`subType`/`status` reflect the most recently seen occurrence.
    - `result`/`error` kept **only when a name occurs exactly once**; dropped on the
      first repeat (`failedCount` still flags failures).
    - Unnamed operations excluded.

  **Exporters (per-store shaping)**
  - CloudWatch (`LambdaLogExporter`, `CloudWatchLogsExporter`) and `DynamoDBExporter`
    emit `operationsByName` **instead of** the array.
  - Array-native exporters (S3/Athena, OpenSearch, Aurora, Redshift) keep the array.
  - `HttpExporter` and `OTelExporter` take an **`operationsFormat`** option
    (`"array"` default | `"by-name"` | `"both"`) since their consumers are arbitrary.

  **VS Code extension (NL→query)**
  - Destination-specific schema handed to the model: `operationsByName` (dot-path) for
    CloudWatch-direct + DynamoDB; JSONB/JSONPath over the array for Aurora; best-effort
    note for the nested `LambdaLogExporter`. Added per-operation-name few-shots +
    README example questions.
  - Removed the stale `PENDING` status from all schema variants.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight-vscode/README.md
+++ b/packages/aws-durable-execution-sdk-js-insight-vscode/README.md
@@ -88,16 +88,23 @@ us.anthropic.claude-sonnet-4-20250514-v1:0
 
 Type a question and click **Ask**. Examples:
 
-| Question                                    | Works best with    |
-| ------------------------------------------- | ------------------ |
-| "show me the last 50 records"               | Any destination    |
-| "count executions by status"                | Any destination    |
-| "show failed executions from the last hour" | Any destination    |
-| "average duration of successful executions" | Aurora, CloudWatch |
-| "failure rate percentage"                   | Aurora             |
-| "average duration grouped by function"      | Aurora             |
-| "show executions longer than 5 seconds"     | Any destination    |
-| "find execution with name abc123"           | DynamoDB, Aurora   |
+| Question                                                | Works best with              |
+| ------------------------------------------------------- | ---------------------------- |
+| "show me the last 50 records"                           | Any destination              |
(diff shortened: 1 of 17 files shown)


yaythomas (approved):
the design doc makes the "array is canonical, map is a per-store projection" decision easy to follow. 🙇 🥇

Issue #674, opened 2026-07-01
[Bug]: Testing SDK Incompatible With V2 Release


PeterBurner:
### Expected Behavior

The `@aws/durable-execution-sdk-js-testing` NPM package should be compatible with `@aws/durable-execution-sdk-js@2.0.0`.

### Actual Behavior

`Durable Execution Testing SDK` cannot be installed together with `Durable Execution SDK` v2. The latest testing package has a peer dependency to `@aws/durable-execution-sdk-js@^1.0.1`.

Error:
```shell
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: undefined@undefined
npm error Found: @aws/durable-execution-sdk-js@2.0.0
npm error node_modules/@aws/durable-execution-sdk-js
npm error   @aws/durable-execution-sdk-js@"2.0.0" from the root project
npm error
npm error Could not resolve dependency:
npm error peer @aws/durable-execution-sdk-js@"^1.0.1" from @aws/durable-execution-sdk-js-testing@1.1.1
npm error node_modules/@aws/durable-execution-sdk-js-testing
npm error   @aws/durable-execution-sdk-js-testing@"1.1.1" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry this command with --force or --legacy-peer-deps to accept an incorrect (and potentially broken) dependency resolution.
```

### Steps to Reproduce

1. Run `npm i @aws/durable-execution-sdk-js@2.0.0 @aws/durable-execution-sdk-js-testing@latest`

### SDK Version

2.0.0

### Node.js Version

24.x

### Is this a regression?

Yes

### Last Working Version

1.1.7

### Additional Context

This version conflict was resolved in commit https://github.com/aws/aws-durable-execution-sdk-js/commit/1e60aee13dd2388fbd1d127bd6a3d3aa1834e28d but never published to NPM.

ParidelPooya:
Thanks for reporting the bug,
We released @aws/durable-execution-sdk-js-testing v1.1.2 that solves the issue:
https://www.npmjs.com/package/@aws/durable-execution-sdk-js-testing/v/1.1.2

PeterBurner:
@ParidelPooya SDK and testing lib are out of sync again. Cannot install SDK v2.2.0 with Testing v1.1.3.
Version was pinned in #687

PR #678, opened 2026-07-02
fix(testing-sdk): bump to 1.1.2 and publish core SDK peer-compat fix

ParidelPooya:
The published `@aws/durable-execution-sdk-js-testing@1.1.1` declares a peer dependency of `@aws/durable-execution-sdk-js@^1.0.1`, which fails to resolve against the core SDK v2.0.0 (issue #674). The source was already widened to `*`, but the version was never bumped, so the fix could not be republished over the existing 1.1.1 on npm.

Changes:
- Bump the testing package version 1.1.1 -> 1.1.2 so the already-correct `*` peer range can be published to the registry.
- Add a regression test (src/__tests__/peer-compatibility.test.ts) that reads both manifests off disk and asserts the core SDK's current version satisfies the declared peer range. Workspace symlinking masks this class of bug, so the test guards against it without needing a registry install.
- Add `semver` and `@types/semver` devDependencies used by the test.

Verified: full testing suite passes (1167 passed); the test fails as expected when the peer range is reverted to `^1.0.1`.

*Issue #, if available:* #674

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5737,6 +5737,13 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/@types/semver": {
+      "version": "7.7.1",
+      "resolved": "https://registry.npmjs.org/@types/semver/-/semver-7.7.1.tgz",
+      "integrity": "sha512-FmgJfu+MOcQ370SD0ev7EI8TlCAfKYU+B4m5T3yXc1CiRN94g/SZPtsCkk506aUDtlMnFZvasDwHHUcZUEaYuA==",
+      "dev": true,
+      "license": "MIT"
+    },
     "node_modules/@types/sinonjs__fake-timers": {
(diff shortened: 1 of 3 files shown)


yaythomas (approved):
1. Core SDK version at PR head is 2.0.0, testing peer range is already "*", testing version bumped 1.1.1 → 1.1.2 in both package.json and package-lock.json. All consistent with the PR description.
2. Issue #674 root cause confirmed: published 1.1.1 shipped peer ^1.0.1, which ERESOLVEs against core 2.0.0. Source was widened to * in a prior commit (1e60aee) but never republished because the version was not bumped. This PR does the bump so the fix can ship.
3. The regression test genuinely guards the bug: satisfies("2.0.0", "^1.0.1") is false
4. semver / @types/semver are added as devDependencies only, not runtime. Correct.

PR #681, opened 2026-07-02
feat(sdk): implement deterministic sampling for Workflow Insight

ParidelPooya:
Implements the previously-stubbed `samplingRate` config for the Workflow
   Insight plugin. Until now the field existed in the types but had no effect
   (the plugin only logged a "not yet implemented" warning). It now controls
   what fraction of executions emit records.

### Behavior
   - `samplingRate` is a number in `[0, 1]`, default `1.0` (every execution).
   - When below `1.0`, only a fraction of executions emit records; sampled-out
     executions schedule **no** records and make **no** exporter calls.
   - The decision is **per-execution and all-or-nothing**: a sampled-in execution
     emits all of its records, a sampled-out execution emits none — records are
     never fragmented.
   - The decision is **deterministic across replays**: it is derived from a hash
     of the execution ARN, which is stable across all invocations/replays of the
     same execution, so a resumed execution never flips its decision mid-flight.
   - Out-of-range or non-numeric values are clamped/defaulted to `1.0` with a
     warning rather than throwing, so a misconfiguration never breaks
     instrumentation.

### Implementation

   - `fnv1a32` — FNV-1a 32-bit hash using `Math.imul` for a correct, cross-engine
     stable 32-bit multiply (a plain `*` loses precision past 2^53).
   - `shouldSampleExecution(executionArn, rate)` — maps the hash into `[0, 1)` and
     samples in when it falls below `rate`.
   - `resolveSamplingRate` — normalizes/clamps the configured rate.
   - The decision is computed once per execution into `ExecutionState.sampledIn`
     and reused; every lifecycle hook is guarded. `wrapInvocation` still runs the
     execution but skips drain/flush when sampled out, and terminal state cleanup
     still runs for sampled-out executions so state entries don't leak.

### Tests

   Adds a `sampling` suite (fully deterministic, no flakiness): rate `0` / `1` /
   omitted, fractional partitioning, reproducibility (identical ARNs → identical
   decisions), stable re-runs, and out-of-range/NaN clamping. Full insight suite:
   29 tests passing.

### Docs

   Updates the `samplingRate` JSDoc, README config section, `plugin-contracts.md`,
   `workflow-insight-design.md`, and marks the sampling tasks/tests done in
   `remaining-tasks.md`.

 
### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight/src/index.ts
+++ b/packages/aws-durable-execution-sdk-js-insight/src/index.ts
@@ -119,21 +119,15 @@ function fnv1a32(input: string): number {
 /**
  * Deterministic, all-or-nothing per-execution sampling decision.
  *
- * The decision must be identical for every invocation and replay of the same
- * execution, otherwise a resumed execution could flip its decision mid-flight
- * and emit fragmented records. We therefore key the hash on the execution's
- * stable identity — its `executionName` — and NOT the full ARN: the ARN carries
- * a per-invocation suffix (`.../<executionName>/<invocationId>`) that changes on
- * replay. Falls back to the raw ARN only when no `executionName` is present.
- *
- * Maps the hash into [0, 1) and samples in when it falls below `rate`.
+ * The execution ARN is stable across all invocations and replays of the same
+ * execution, so hashing it directly yields a decision that never changes
+ * mid-flight — a resumed execution always reaches the same conclusion. Maps the
+ * hash into [0, 1) and samples in when it falls below `rate`.
  */
 function shouldSampleExecution(executionArn: string, rate: number): boolean {
   if (rate >= 1) return true;
   if (rate <= 0) return false;
-  const { executionName } = parseExecutionArn(executionArn);
-  const key = executionName || executionArn;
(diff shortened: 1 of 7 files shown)


packages/aws-durable-execution-sdk-js-insight/src/index.ts
```diff
@@ -375,6 +424,12 @@ export function workflowInsight(
     startTime: Date;
     parsedArn: ParsedArn;
     cachedInput: unknown;
+    /**
+     * Deterministic sampling decision for this execution. Computed once from the
+     * execution's stable identity so every invocation/replay agrees, then reused
+     * by all hooks to skip work entirely when the execution is sampled out.
+     */
+    sampledIn: boolean;
```
SilanHe, line 432: nit: does this mean sampledInvocation? this is a little confusing. `isSampled` is probably better

SilanHe (approved):

packages/aws-durable-execution-sdk-js-insight/src/index.ts
ParidelPooya: No, it stands for:
this execution was selected (sampled in) to emit records.

PR #683, opened 2026-07-03
feat(sdk): enforce maxRecordSizeBytes with best-effort record truncation

ParidelPooya:
Implements the previously-unused `InsightExporter.maxRecordSizeBytes` for the
   Workflow Insight plugin. Until now the field existed on the interface but was
   never read, so oversized records were emitted as-is and could be rejected by
   size-capped destinations (CloudWatch Logs 256 KB, DynamoDB 400 KB, etc.).

   When a record's serialized JSON exceeds an exporter's limit, the plugin now
   truncates a **per-exporter copy** before export — so the same execution can be
   emitted in full to one destination and trimmed to another. The shared record is
   never mutated.

### Drop order (best-effort, least-destructive first)

   Applied until the record fits the limit:

   1. operation `result` fields, **oldest operation first** (the operation is kept
      and marked `truncated: true`);
   2. whole operations, **oldest first**;
   3. last resort — execution `input`, then `output`.

   Identity/timeline fields are never dropped. Because `content.input` /
   `content.output` transforms run *before* truncation, they remain the preferred
   way to bound large payloads.

### Markers ("cut, not missing")

   Additive, optional fields — `schemaVersion` stays `"1.0"`:

   - record `truncated: boolean` — anything was cut
   - `OperationRecord.truncated: boolean` — that operation's result was dropped
   - `droppedOperations: number` — whole operations removed
   - `droppedInput` / `droppedOutput: boolean` — last-resort payload drops

### Per-exporter defaults (all overridable)
```
   | Exporter(s)                                    | Default   |
   | ---------------------------------------------- | --------- |
   | Lambda log, CloudWatch Logs, SQS, EventBridge  | 256 KB    |
   | DynamoDB                                       | 400 KB    |
   | Aurora, Redshift, Firehose, OTel               | 1 MB      |
   | S3                                             | 5 MB      |
   | OpenSearch                                     | 10 MB     |
   | HTTP, File, Timestream                         | none¹     |
```
   ¹ No default — truncation disabled unless `maxRecordSizeBytes` is set.

   ### Docs

   `README.md` (new "Record size & truncation" section + record fields + custom
   exporter example), `plugin-contracts.md`, `workflow-insight-design.md`, and
   `remaining-tasks.md` updated and reconciled.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight/src/exporters/aurora-exporter.ts
+++ b/packages/aws-durable-execution-sdk-js-insight/src/exporters/aurora-exporter.ts
@@ -36,6 +36,12 @@ export interface AuroraExporterConfig {
 
   /** AWS region. If omitted, uses the SDK default. */
   region?: string;
+
+  /**
+   * Max serialized record size before truncation.
+   * Default: 1_000_000 (Aurora Data API practical row/statement limit).
+   */
+  maxRecordSizeBytes?: number;
 }
 
 /**
@@ -56,6 +62,7 @@ export class AuroraExporter implements InsightExporter {
   private readonly table: string;
   private readonly engine: "postgresql" | "mysql";
   private readonly client: RDSDataClient;
+  readonly maxRecordSizeBytes: number;
 
   constructor(config: AuroraExporterConfig) {
     this.resourceArn = config.resourceArn;
@@ -63,6 +70,7 @@ export class AuroraExporter implements InsightExporter {
     this.database = config.database;
     this.table = sanitizeIdentifier(config.table ?? "workflow_insight");
     this.engine = config.engine;
+    this.maxRecordSizeBytes = config.maxRecordSizeBytes ?? 1_000_000;
     this.client = new RDSDataClient(
       config.region ? { region: config.region } : {},
     );
--- a/packages/aws-durable-execution-sdk-js-insight/src/index.ts
+++ b/packages/aws-durable-execution-sdk-js-insight/src/index.ts
@@ -327,8 +327,12 @@ export class LambdaLogExporter implements InsightExporter {
     this.maxRecordSizeBytes = config.maxRecordSizeBytes ?? 256_000;
   }
 
+  render(record: WorkflowInsightRecord): unknown {
+    return withOperationsByName(record);
+  }
+
   async export(record: WorkflowInsightRecord): Promise<void> {
-    console.log(JSON.stringify(withOperationsByName(record)));
+    console.log(JSON.stringify(this.render(record)));
   }
 }
 
@@ -382,11 +386,16 @@ class ExportScheduler {
         this.pending = undefined;
         // allSettled so one failing/slow exporter never blocks or fails the others,
         // and an export error never propagates into the execution. Each exporter
-        // gets a copy truncated to its own maxRecordSizeBytes (no-op when unset).
+        // gets a copy truncated to its own maxRecordSizeBytes (no-op when unset),
+        // measured against the exact shape that exporter emits (its `render`).
         await Promise.allSettled(
           this.exporters.map((exporter) =>
             exporter.export(
-              truncateRecord(record, exporter.maxRecordSizeBytes),
+              truncateRecord(
+                record,
+                exporter.maxRecordSizeBytes,
+                exporter.render ? (r) => exporter.render!(r) : undefined,
+              ),
             ),
           ),
         );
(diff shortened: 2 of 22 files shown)


packages/aws-durable-execution-sdk-js-insight/src/index.ts
```diff
@@ -317,8 +320,19 @@ function buildOperationRecords(
  * @experimental This class is experimental and may change in future releases.
  */
 export class LambdaLogExporter implements InsightExporter {
```
SilanHe, line 322: nit: if this is exporting the logs to CloudWatch logs, should we just use the cloudwatch-logs-exporter.ts instead where LambdaLogExporter is used?

packages/aws-durable-execution-sdk-js-insight/src/truncation.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+    return out;
+  };
+
+  const fits = (): boolean => {
+    const size = jsonByteSize(render(candidate()));
+    return size !== undefined && size <= maxBytes;
+  };
+
+  // Phase 1: drop operation results, oldest first. The operation is kept and
```
SilanHe, line 102: nit: there could be cases where result:myResult is smaller than truncated:true. Could do a micro optimization here where we don't drop the result if the metadata+result is smaller than the new field we're introducing

packages/aws-durable-execution-sdk-js-insight/src/exporters/aurora-exporter.ts
```diff
@@ -56,13 +62,15 @@ export class AuroraExporter implements InsightExporter {
...
   private readonly client: RDSDataClient;
+  readonly maxRecordSizeBytes: number;
 
   constructor(config: AuroraExporterConfig) {
     this.resourceArn = config.resourceArn;
     this.secretArn = config.secretArn;
     this.database = config.database;
     this.table = sanitizeIdentifier(config.table ?? "workflow_insight");
     this.engine = config.engine;
+    this.maxRecordSizeBytes = config.maxRecordSizeBytes ?? 1_000_000;
```
SilanHe, line 73: nit: how come we don't define render for some `exporters`? repeats

packages/aws-durable-execution-sdk-js-insight/src/index.ts
```diff
@@ -371,9 +385,19 @@ class ExportScheduler {
...
+        // gets a copy truncated to its own maxRecordSizeBytes (no-op when unset),
+        // measured against the exact shape that exporter emits (its `render`).
         await Promise.allSettled(
-          this.exporters.map((exporter) => exporter.export(record)),
+          this.exporters.map((exporter) =>
+            exporter.export(
+              truncateRecord(
+                record,
+                exporter.maxRecordSizeBytes,
+                exporter.render ? (r) => exporter.render!(r) : undefined,
```
SilanHe: looking at the `truncateRecord` function, it looks like it's directly calling `render` within it's body without any checks. If the `render` is undefined, will the truncateRecord code throw?

ParidelPooya: No, it won't throw. truncateRecord's render parameter has a default (= (r) => r), and in JS/TS, default parameter values apply whenever the argument is undefined — whether omitted or
  passed explicitly. 

  That said, your instinct was reasonable — the old call site (exporter.render ? (r) => exporter.render!(r) : undefined) made you reason through the default-parameter rule to be sure it was safe, and had
  a redundant non-null assertion. 
  
  Simplified it to exporter.render?.bind(exporter), which is clearer and equivalent, and added a test that explicitly passes undefined and asserts no throw + identical behavior to
  omitting the arg — so the contract is now verified, not just documented.

packages/aws-durable-execution-sdk-js-insight/src/exporters/aurora-exporter.ts
ParidelPooya: Deduplicated in new commit!

packages/aws-durable-execution-sdk-js-insight/src/index.ts
ParidelPooya: They solve different problems, so I'd keep both: LambdaLogExporter piggybacks on Lambda's existing stdout→CloudWatch pipe (writes to the function's own log group, zero IAM/config needed) and is meant to be the
   zero-setup default. CloudWatchLogsExporter is the opt-in exporter for routing to a dedicated log group of the customer's choosing, with its own stream partitioning — it needs logGroupName config and
   logs:CreateLogStream/logs:PutLogEvents IAM permissions. CloudWatchLogsExporter's doc comment already calls out this distinction explicitly. Collapsing them would force the default zero-config path to require IAM
   setup it doesn't need today.

SilanHe (approved):

packages/aws-durable-execution-sdk-js-insight/src/index.ts
SilanHe: ah oops, I misread the default value. I also understand why some exporters don't define `render` then.

PR #686, opened 2026-07-03
feat(sdk): add SQS live view to the Workflow Insight VS Code extension

ParidelPooya:
SQS has no query engine, so it doesn't fit the extension's existing "ask a
   question → LLM generates a query → run it → render a table" pipeline used by
   CloudWatch Logs, DynamoDB, and Aurora. This adds `sqs` as a distinct
   destination type with its own live view instead: start a listener, watch
   messages arrive in real time, click a row for full detail.

### What's new

   **Settings**
   - `workflowInsight.destinationType` gains `"sqs"`.
   - New `workflowInsight.sqsQueueUrl` and `workflowInsight.sqsDeleteAfterRead`
     (boolean, default `false`).

   **Live listener (extension host)**
   - `src/sqs.ts`: long-polls the queue (`ReceiveMessageCommand`,
     `WaitTimeSeconds=20`) in a loop until aborted.
   - **Peek-only by default** — other consumers still receive every message.
     Enabling `sqsDeleteAfterRead` makes the Explorer a real consumer
     (`DeleteMessageBatchCommand` after each batch).
   - New `startListening`/`stopListening` messages; an `AbortController` tracks
     the active listener and is aborted on both `stopListening` and panel
     dispose, so closing the Explorer can't leak a running poll loop.

   **Webview**
   - `SqsLiveView.tsx`: Start/Stop control, running message count, JSON message
     bodies parsed into the same columns/rows shape the other destinations use
     (so results look consistent regardless of source) — falls back to raw
     display if a body isn't valid JSON.
   - Table shows only the primary fields (status, functionName, executionName,
     durationMs, startTime, endTime); the rest of each record is one click away.
   - `ResultsTable` gets a single-select radio column (previously rows were
     clickable with no visual indicator) and a `RecordDetail` modal:
     - `operations`/`operationsByName` render as their own selectable table
       instead of a raw JSON blob; selecting an operation shows its
       `result`/`error` as syntax-highlighted JSON.
     - `input`/`output`/`error` each get their own tab via Cloudscape's
       `CodeView` (`@cloudscape-design/code-view`) with JSON highlighting,
       instead of an inline stringified string.
   - Visualize (chart) flow now also works for the SQS view, reusing the
     existing `VisualizePage`.

   ### Testing

   - `tsc --noEmit` clean for both the extension and webview-ui packages.
   - Full extension build (esbuild + webview bundle) succeeds.
   - Manually verified end-to-end against a real deployed SQS queue: started
     listening, confirmed live message counts, parsed record columns, operation
     detail (result/error), input/output JSON views, and the Visualize chart.
   - No existing test suite for this package.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight-vscode/webview-ui/src/App.tsx
+++ b/packages/aws-durable-execution-sdk-js-insight-vscode/webview-ui/src/App.tsx
@@ -8,7 +8,8 @@ import { QueryPanel } from "./QueryPanel";
 import { ResultsTable } from "./ResultsTable";
 import { VisualizePage } from "./VisualizePage";
 import { SettingsModal } from "./SettingsModal";
-import type { InboundMessage, Settings } from "./types";
+import { SqsLiveView } from "./SqsLiveView";
+import type { InboundMessage, Settings, SqsMessageRow } from "./types";
 import { DEFAULT_SETTINGS } from "./types";
 
 applyMode(Mode.Dark);
@@ -26,6 +27,8 @@ export function App() {
   const [downloadPercent, setDownloadPercent] = useState(0);
   const [page, setPage] = useState<Page>("data");
   const [explanation, setExplanation] = useState("");
+  const [sqsMessages, setSqsMessages] = useState<SqsMessageRow[]>([]);
+  const [sqsListening, setSqsListening] = useState(false);
 
   const handleMessage = useCallback((event: MessageEvent<InboundMessage>) => {
     const msg = event.data;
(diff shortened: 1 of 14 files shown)


packages/aws-durable-execution-sdk-js-insight-vscode/webview-ui/src/App.tsx
```diff
@@ -56,6 +59,18 @@ export function App() {
       case "settingsSaved":
         setSettingsOpen(false);
         break;
+      case "sqsStatus":
+        setSqsListening(msg.listening);
+        break;
+      case "sqsMessages":
+        setSqsMessages((prev) => {
```
SilanHe, line 66: should we limit the max number of SqsMessages we keep in memory? I think we do trim the messages for display below but not the ones tracked by the underlying sqsMessages.

SilanHe (commented):
I haven't ever worked with React code in production so I can't reasonably review that portion of the code. If you're seeking a second pair of eyes on that portion of the code, you'll have to ask someone else.

packages/aws-durable-execution-sdk-js-insight-vscode/webview-ui/src/App.tsx
ParidelPooya: Good catch — sqsMessages state was growing unbounded across a listening session; only the render/visualize call sites were slicing to MAX_DISPLAYED_MESSAGES, not the underlying state itself. Fixed in new commit:
   the handler now trims to the last MAX_DISPLAYED_MESSAGES right after appending each de-duped batch, so state can't grow past that regardless of how long the listener runs. This also avoids the seen dedupe Set
   getting rebuilt from an ever-growing array on every batch.

SilanHe (approved):

PR #689, opened 2026-07-03
feat(sdk): add S3 + Athena support to Workflow Insight Explorer

ParidelPooya:
Adds S3 + Athena as a fifth queryable destination in the Workflow Insight
  Explorer VS Code extension, alongside CloudWatch Logs, DynamoDB, Aurora,
  and the SQS live view — completing the destination this extension's
  original design doc called for.

  - **Query engine**: Trino/Presto SQL via Amazon Athena, with a new
    StartQueryExecution → poll → paginated GetQueryResults runner mirroring
    the existing CloudWatch Logs Insights async-poll pattern.
  - **Zero manual setup**: the Glue database/table (JSON SerDe, Hive
    year=/month=/day= partitioning matching S3Exporter's exact output) are
    provisioned directly in CDK for the example stack, and auto-created on
    Settings save for customers pointing at their own S3Exporter buckets.
  - **NL→SQL prompt**: new schema/dialect/few-shots teaching the model
    Athena's dialect, `UNNEST(operations)` for per-operation queries, and
    Athena's JSON-key-lowercasing behavior for `input`/`output` field access
    — including a rule + few-shots specifically preventing the model from
    treating a JSON field (e.g. `claimType`) as a bare column in
    `GROUP BY`/`ORDER BY`/`WHERE`.
  - **Row-detail drill-down for every destination** (not just S3+Athena):
    fixed a pre-existing gap where Ask results only ever showed the columns
    the generated query selected, with no way to see `operations`/`input`/
    `output`. Row-level results now get a stable per-execution identifier
    injected into the query if missing, and clicking a row fetches the full
    record on demand — skipped entirely for aggregate results (`COUNT`,
    `GROUP BY`, etc.), where there's no single execution a row corresponds

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16200,9 +16200,12 @@
         "node-llama-cpp": "^3.18.1"
       },
       "devDependencies": {
+        "@types/jest": "^29.5.0",
         "@types/node": "^20.0.0",
         "@types/vscode": "^1.90.0",
         "esbuild": "^0.24.0",
+        "jest": "^29.7.0",
+        "ts-jest": "^29.2.0",
         "typescript": "~5.8.0"
       },
       "engines": {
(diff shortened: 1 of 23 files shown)


yaythomas (approved):
webview-ui/src/SettingsModal.tsx The Athena settings help text still says:

"...creates it ... and runs MSCK REPAIR TABLE to discover existing partitions."

The implementation deliberately does not do this, athena.ts uses partition projection, and your own comments there state MSCK REPAIR is a no-op/disallowed on a projection-enabled table. 

src/extension.ts, onEnsureAthenaTable Docstring: "Auto-create (or verify) the Glue table backing Athena queries, and discover any Hive partitions S3Exporter has already written." The method no longer discovers partitions (projection does that at query time).

PR #690, opened 2026-07-04
feat(sdk): conversational agentic assistant for the Workflow Insight extension

ParidelPooya:
Adds an agentic, conversational assistant to the Workflow Insight VS Code
  extension. Instead of turning one question into one query, it explores the data
  across multiple steps, answers in plain English, and holds a continuing
  conversation — with each answer's result table kept inline in the thread.

  This began as an opt-in "advanced mode" beside the original single-shot
  behavior, but per review that split was collapsed: the assistant is **always**
  agentic now, so there's a single code path and UX to maintain.

  ## What's new

  **Agentic loop (all providers, two implementations)**
  - **Bedrock** runs a multi-step tool loop (`agentLoop.ts`) over Converse:
    `run_query` to explore, `run_javascript` to compute, `finish` to answer —
    "explore first, then answer," with cross-turn conversation memory.
  - **Copilot / local** run a verify → refine loop: generate a query, run it,
    judge whether it answers the question, refine, and return a prose answer.
    Conversation context is threaded through every attempt (initial + refines).
    (Native multi-step tool use for these providers is planned — see Roadmap.)
  - `run_javascript` runs in a real WASM sandbox (QuickJS): no
    filesystem/network/host access, with time + memory limits. It computes over
    the full result set (not the model's display sample); a truncation note is
    returned past a safety cap so aggregates are never silently partial.
  - Works across DynamoDB, Aurora, S3/Athena, and CloudWatch Logs.

  **Conversational UX**
  - Persistent multi-turn conversation: prose answers, prior turns as context,
    and a "New session" control.
  - Each turn keeps its own result table + collapsible agent-steps transcript
    inline (question → answer → table, repeated down the thread).
  - Chat-style inline composer at the bottom; Enter sends, Shift+Enter newline.
  - Result tables paginate small (5/page) so they don't bury the conversation.
  - Answers summarize results (counts, ranges, the specific insight) instead of
    re-dumping the rows already shown in the table.
  - Conceptual/metadata answers ("which field holds the amount?") show prose
    only — no unrelated table is rendered.

  **Configurable local model**
  - New `workflowInsight.localModel` setting with vetted presets, defaulting to
    the best tool-calling model that runs on a laptop: **Llama-3-Groq-8B
    Tool-Use** (~89% on the Berkeley Function-Calling Leaderboard, ~4.9 GB).
    Alternatives: Phi-3.5-mini (~2.4 GB), Qwen2.5-Coder-3B (~2.2 GB).

  **Guards, cost & correctness**
  - `workflowInsight.agenticMaxIterations` — the single, source-agnostic
    cost/effort cap; plus an oscillation guard (repeated-query detection).
  - The finish query, if already run during exploration, is reused instead of
    re-scanned (per-question result cache keyed by executed SQL) — avoids
    double-billing Athena scans.
  - Row drill-down is gated on query shape (deterministic), and identifier
    injection safely skips DISTINCT/aggregate/set-operator/derived queries — so
    "click a row for the full record" works consistently across providers
    without corrupting queries.
  - Athena/Trino prompt fixes: JSON-key enumeration, correct `UNNEST` aliasing
    (no more `t.op.name` → `COLUMN_NOT_FOUND`), DISTINCT/ORDER BY guidance.
  - Bedrock output-token caps raised (finish/answer/query won't truncate).

  ## Removed
  - `workflowInsight.agenticMode` (basic mode) — always agentic now.
  - `workflowInsight.agenticMaxScannedMB` (Athena-only scan budget) — it was a
    no-op for every non-Athena source; the iteration cap is the universal guard,
    and per-query Athena scan limits belong in the workgroup's
    `bytes_scanned_cutoff_per_query` (documented in the setting + README).
  - The dead `rowLevel` model flag (drill-down is now decided by query shape).

  ## New files
  `src/agentLoop.ts`, `src/sandbox.ts`, `src/verdict.ts`,
  `webview-ui/src/AgentTranscript.tsx`, plus `src/sandbox.test.ts`,
  `src/verdict.test.ts`.

  ## Testing
  - Unit tests for the sandbox (incl. the security boundary + repeated calls),
    verdict/prompt helpers, and query-shape/identifier-injection logic —
    56 passing.
  - Extension-host + webview typechecks and full build clean.
  - Agent loop, prose answers, conversation continuity, `run_javascript`,
    drill-down injection, result summarization, and the query-omitted-conceptual
    path verified live against Bedrock + Athena (account 730758745077,
    us-east-1). New local-model GGUF URLs verified resolvable (not download-run
    in CI).

  ## Roadmap / follow-ups
  - Native multi-step agentic loop for Copilot and local providers (a
    provider-agnostic loop with per-provider adapters), so all three get
    explore-then-answer + `run_javascript` + cross-turn memory, not just Bedrock.
    The configurable local model added here is the setup for that.

  ## Notes / scope
  - Today the full explore-then-answer + `run_javascript` + cross-turn memory
    require Bedrock; Copilot/local use the verify/refine loop with prose answers
    and threaded conversation context.
  - Conversation tables are webview session state (a reload / New session clears
    them); backend memory threads the summarized Q&A text, not the tables.
  - The local-model default download is ~4.9 GB (a smaller preset is selectable).


packages/aws-durable-execution-sdk-js-insight-vscode/src/agentLoop.ts
```diff
@@ -0,0 +1,530 @@
...
+      .filter((t): t is ToolUseBlock => !!t);
+
+    if (toolUses.length === 0) {
+      // No tool call — treat any prose as a final answer, else stop.
+      const text = blocks
+        .map((b) => ("text" in b && b.text ? b.text : ""))
+        .join("")
+        .trim();
+      if (text) {
+        return { query: lastGoodQuery ?? "", explanation: "", answer: text };
```
yaythomas, line 445: the "finish" path guards against this, the comment there explains it falls back to "" on purpose so a conceptual answer shows no table?

yaythomas (approved):

packages/aws-durable-execution-sdk-js-insight-vscode/src/agentLoop.ts
ParidelPooya: Yes, no table when we can answer the customer question in sentence.

PR #691, opened 2026-07-06
feat(sdk): LLM-driven chart configuration on the Visualize page

ParidelPooya:
Builds on the conversational Workflow Insight assistant (#690). The Visualize page previously configured charts with a local keyword heuristic that guessed axes from column order and, for the free-text box, only detected the chart TYPE and discarded the fields the user named — so a request like "record_count as data, product_category and hour_bucket as axis, stacked bar" rendered with the wrong encoding. It also hid every chart type the model didn't suggest.

Now the model configures the chart:

- New llm.ts generateChartSpec turns a request (a chosen chart type and/or a free-text description) plus the result's column names and which are numeric into a validated Vega-Lite {mark, encoding} — choosing the field for each channel and optional styling (axis scale/min, color scheme, sort, titles). It dispatches across bedrock/copilot/local via a shared completeText helper and throws on a missing/invalid spec so failures surface instead of rendering a wrong chart. Only column names/types and the request are sent — never rows.
- extension.ts handles a new "visualize" webview message (onVisualize) and replies with "chartSpec" / "chartSpecError".
- VisualizePage: the preset button row is now a dropdown listing ALL chart types with the model-suggested ones tagged "Recommended"; both selecting a type and submitting free text call the model (loading state, inline errors, Enter-to-submit). The local applyPreset/column-guessing heuristic is removed.

Verified live against Bedrock: the example request yields record_count on y (stack:zero), product_category on x, hour_bucket as color; selecting "stacked-bar" with no text picks sensible channels and styling. Typecheck (extension + webview) and build are clean; 98 tests pass.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight-vscode/src/chartSpec.test.ts
+++ b/packages/aws-durable-execution-sdk-js-insight-vscode/src/chartSpec.test.ts
@@ -105,6 +105,12 @@ describe("parseChartSpec", () => {
     ).toThrow(/malformed encoding/i);
   });
 
+  it("rejects a dynamic expr/signal in the encoding", () => {
+    const raw =
+      '{"mark":"bar","encoding":{"x":{"field":"hour_bucket"},"y":{"value":{"expr":"width"}}}}';
+    expect(() => parseChartSpec(raw, COLS)).toThrow(/dynamic expression/i);
+  });
+
   it("rejects disallowed top-level keys (data/transform/etc.)", () => {
     const withData =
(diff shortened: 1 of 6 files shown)


yaythomas (approved):
verified the defense-in-depth story holds: the webview CSP in extension.ts is default-src 'none'; connect-src data: blob:; img-src {cspSource} data: blob: — no http(s), so even a spec that slipped a data.url or an image mark past the validator can't fetch anything. The validator's top-level allowlist + mark allowlist + whole-spec expr/signal scan is a solid second layer.

PR #692, opened 2026-07-06
feat(sdk): emit operationName in default logger output

ParidelPooya:
The default logger emitted requestId, executionArn, tenantId, operationId, and attempt, but never operationName. This diverged from the documented log format and from the Python and Java SDKs, which both emit operationName.

Thread the operation/step/child-context name through the async-local context tracker (runWithContext) and surface it via getDurableLogData so the default logger includes operationName when available. It is derived from the operation name (like the Java SDK) and is omitted for the root context and when no name is present, keeping the change backward-compatible.

- context-tracker: add operationName to ContextInfo and runWithContext
- step, wait-for-condition, and run-in-child-context handlers pass the name
- DurableLogData: add optional operationName field
- durable-context: populate operationName in getDurableLogData (omit at root)
- default-logger: emit operationName when present
- add unit/integration tests; update runWithContext argument assertions

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/src/examples/logger-test/log-levels/logger-log-levels.test.ts
+++ b/packages/aws-durable-execution-sdk-js-examples/src/examples/logger-test/log-levels/logger-log-levels.test.ts
@@ -51,6 +51,7 @@ createTests({
             executionArn: expect.any(String),
             ...(hasOperationId && {
               operationId: expect.any(String),
+              operationName: expect.any(String),
             }),
             ...(attemptCount && {
               attempt: attemptCount,
@@ -165,6 +166,7 @@ createTests({
               level: "ERROR",
               executionArn: expect.any(String),
               operationId: expect.any(String),
+              operationName: expect.any(String),
               message: expect.stringMatching(
                 /^Error from child context with Error object: Error: Child context error/,
               ),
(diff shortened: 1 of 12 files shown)


yaythomas (dismissed):
- All production call sites of runWithContext are covered. There are 5 in src/ (not counting tests): step-handler.ts:326, wait-for-condition-handler.ts:255, and two in run-in-child-context-handler.ts (the ReplayChildren replay path at :260 and executeChildContext at :384) all now pass the operation name. The fifth, with-durable-execution.ts:123, is the root context ("root", no name) and is correctly left untouched. No operation path that runs customer code in a tracked context was missed.
- operationName is the human-readable name, operationId stays hashed (hashId(contextId)). Good — they carry different information and the test (operationName: "fetch-user" alongside a hashed operationId) confirms it.
- Root is doubly-guarded. getDurableLogData in packages/aws-durable-execution-sdk-js/src/context/durable-context/durable-context.ts requires contextId !== "root" before emitting, and the root call site passes no name anyway.
- Field ordering matches the order-sensitive tests. formatDurableLogData in packages/aws-durable-execution-sdk-js/src/utils/logger/default-logger.ts inserts operationName between operationId and attempt, matching both the DurableLogData interface order and the exact JSON.stringify assertions in default-logger.test.ts.
- Typechecks cleanly. DefaultDurableLogEntry extends DurableLogData, so adding the optional field to packages/aws-durable-execution-sdk-js/src/types/logger.ts is enough for result.operationName = … to type.

yaythomas (approved):

PR #693, opened 2026-07-06
feat(sdk): add updated operations to invocation info

SilanHe:
*Issue #, if available:*

*Description of changes:*

InvocationInfo now has an updatedOperations: Record<string, OperationInfo> field — the subset of operations that changed externally between invocations (populated from event.UpdatedOperationIds)
Type hierarchy (before → after):

BEFORE:

```
InvocationBaseInfo: { requestId, executionArn }
InvocationInfo extends InvocationBaseInfo: { isFirstInvocation, executionInput, operations }
InvocationEndInfo extends InvocationInfo: { status, executionResult?, executionError? }
OperationChangeInfo extends InvocationBaseInfo: { updatedOperations, operations }
```
AFTER:

```
InvocationBaseInfo: { requestId, executionArn, executionInput, operations }
InvocationInfo extends InvocationBaseInfo: { isFirstInvocation, updatedOperations }
InvocationEndInfo extends InvocationBaseInfo: { status, executionResult?, executionError? }
OperationChangeInfo (standalone): { executionArn, updatedOperations, operations }
```

Key design decisions:
InvocationEndInfo no longer carries isFirstInvocation or updatedOperations — no onInvocationEnd consumer used them
OperationChangeInfo is fully decoupled — it only needs executionArn, not requestId
executionInput and operations live on InvocationBaseInfo so both InvocationInfo and InvocationEndInfo get them without duplication


packages/aws-durable-execution-sdk-js/src/types/plugin.ts
```diff
@@ -139,19 +149,19 @@ export interface InvocationInfo extends InvocationBaseInfo {
...
 }
 
 /**
  * Information provided when operations change during execution.
  *
  * @experimental This interface is experimental and may be changed or removed in future releases.
  */
-export interface OperationChangeInfo extends InvocationBaseInfo {
+export interface OperationChangeInfo {
+  executionArn: string;
```
zhongkechen, line 164: executionArn or durableExecutionArn?

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
```diff
@@ -129,8 +131,16 @@ export interface InvocationBaseInfo {
...
+  /**
+   * Operations that were updated externally between the previous invocation and this one
+   * (e.g., a wait timer expired, a callback was received, or a chained invoke completed).
+   *
+   * This is a subset of `operations` containing only those whose status changed since the
+   * last checkpoint. On the first invocation this will be an empty record.
+   *
+   * Populated from the `UpdatedOperationIds` field in the durable Lambda invocation input.
+   */
+  updatedOperations: Record<string, OperationInfo>;
```
zhongkechen, line 143: is this useful to plugins? I suppose they should only care whether an operation is updated and shouldn't care it's updated whether externally between invocations or internally in an invocation.

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
SilanHe: I thought the conclusion was that we were using executionArn going forward?

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
SilanHe: I spoke with Pooya offline about making this change, I think he has a use for it

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
```diff
@@ -119,6 +119,8 @@ export interface AttemptEndInfo extends AttemptInfo {
 export interface InvocationBaseInfo {
   requestId: string;
   executionArn: string;
```
SilanHe, line 121: executionArn here

packages/aws-durable-execution-sdk-js/src/types/plugin.ts
zhongkechen: Only logging metadata is using executionArn for now. Everywhere else is using durableExecutionArn.

SilanHe: Can I change this in a follow up? I think this requires me to change the insights plugin as well. In the current PR, I'm merely keeping the status quo

zhongkechen (approved):

PR #706, opened 2026-07-09
feat(sdk): Explorer query/ask/agent modes, saved queries + export, local-server LLM, and AI consent

ParidelPooya:
Enhancements to the Workflow Insight Explorer VS Code extension: three explicit
   query modes, saved queries and result export, an OpenAI-compatible local-server
   LLM provider, `.vsix` packaging for preview distribution, and an AI-usage
   consent + data-handling disclosure across every surface.

   ### What's included

   **Composer modes** (Send is now a split dropdown; the choice is persisted in
   `workflowInsight.queryMode` and restored on open):
   - **Query** — runs the text you type verbatim as a read-only query (no LLM, no
     drill-down injection); still read-only-enforced and row-capped. Prefills a
     destination-appropriate starter query and posts an explicit row-count summary.
   - **Ask** — one NL→query translation, run once (no agent loop).
   - **Agent** — the existing agentic explore→answer loop.
   - Distinct per-mode Send icon (query=script, ask=search, agent=gen-ai).

   **Saved queries & export** — each result table has an Actions menu: Export as
   CSV, Export as JSON (saved via the host), and Save query to favorites. Favorites
   persist in globalState and can be re-run from an icon-only picker in the composer
   (filtered to the current destination).

   **Local LLM support** — new `local-server` provider that talks to an
   OpenAI-compatible endpoint you run (Ollama, LM Studio, llama.cpp), configured via
   `workflowInsight.localServerUrl` / `localServerModel`. No native module, so it
   works from the packaged `.vsix` on every platform.

   **AI consent & data handling** — before the first use of any AI feature
   (Ask/Agent/Visualize), a disclosure modal explains that an LLM is used and that
   the request text plus limited data (result column names + a small sample of rows)
   are sent to the selected provider; Query mode sends nothing. Consent is stored in
   `workflowInsight.aiDisclosureAcceptedVersion` and versioned so wording changes
   re-prompt. The disclosure — including per-provider data-flow (Bedrock→AWS,
   Copilot→GitHub, local-server→your endpoint, on-device→never leaves the machine) —
   appears in the consent modal, the in-app settings panel, the native VS Code
   settings descriptions, and the README.

   **Reliability fix** — the Bedrock agent loop now shows the last row-bearing query
   result when the model finishes with a prose-only answer, so a follow-up data
   question still renders its table.

   **Packaging / distribution** — `npm run package` builds a `.vsix` into `vsix/`
   (git-ignored); preview builds are shared via GitHub Releases. README documents
   install-from-VSIX and provider setup.

   ### ⚠️ Legal review required

   The AI-usage disclosure wording (consent modal, settings descriptions, README)
   is a **placeholder pending Legal review** — see `AiConsentModal.tsx` and
   `AI_DISCLOSURE_VERSION` in `webview-ui/src/types.ts`. When approved copy lands,
   update the text and bump `AI_DISCLOSURE_VERSION` to re-prompt consented users.


yaythomas (approved):
:shipit:

PR #708, opened 2026-07-09
ci: use SAM for integration test deployments

zhongkechen:
## Summary
- deploy regular integration-test functions through generated SAM stacks instead of per-function Lambda API calls
- update example/SAM docs for the new workflow

image

## Validation
- node --check .github/workflows/scripts/integration-test/integration-test.js
- git diff --check
- sam validate -t packages/aws-durable-execution-sdk-js-examples/template.yml --region us-west-2


.github/workflows/integration-tests.yml
```diff
@@ -7,10 +7,10 @@ on:
         required: true
         type: string
 
-# Cancel when pull request is updated for this node version
+# Serialize updates to this PR/runtime stack.
 concurrency:
-  group: ${{ github.head_ref || github.ref_name || github.run_id }}-${{ inputs.node-version }}-integ
-  cancel-in-progress: true
+  group: pr-${{ github.event.pull_request.number || github.event.number || github.head_ref || github.ref_name || github.run_id }}-${{ inputs.node-version }}-integ
```
SilanHe: are we not going to restrict this similarly to in Python and Java integration tests?

https://github.com/aws/aws-durable-execution-sdk-java/blob/main/.github/workflows/e2e-tests.yml#L27

does this mean we're keeping the same model for JS where we're deploying one integration test stack per PR?

zhongkechen: I tried to keep this PR small and kept the existing code and behavior as much as possible, e.g. not only the per-pr stack model, but also the existing JS script to run multiple steps (instead of workflow steps). I can create separate PRs to follow up.

zhongkechen: Never mind. I've updated this PR with the per-runtime change.

packages/aws-durable-execution-sdk-js-examples/scripts/README.md
```diff
@@ -2,7 +2,7 @@
 
 This directory contains build and automation scripts for the DurableExecutionsTypeScriptExamples package.
```
SilanHe, line 3: nit: this was missed by me initially, but we should probably remove this. We can do this in a follow up

.github/workflows/cleanup-integration-tests.yml
```diff
@@ -1,98 +0,0 @@
-name: Cleanup Integration Tests
```
SilanHe, line 1: nit: description says 

> add daily cleanup for PR testing stacks older than 7 days

I remember that we do delete this testing file but I thought you had included another cleanup script?

SilanHe (approved):

packages/aws-durable-execution-sdk-js-examples/scripts/README.md
zhongkechen: I can follow up

.github/workflows/cleanup-integration-tests.yml
zhongkechen: I think you meant this one I added earlier and removed in the final version of this PR after we switch to one stack per runtime model:
https://github.com/aws/aws-durable-execution-sdk-js/pull/708/changes/14163c68a2a089a8a249b0d47ea752aa5ea0937b

Removed this from PR description

PR #711, opened 2026-07-10
feat(sdk): return a completion decision from shouldComplete

ParidelPooya:
Align the map/parallel `shouldComplete` API with the Java SDK so a custom predicate can declare whether early completion is a success or a failure, and so it can no longer be combined with the threshold fields.

- `CompletionConfig` is now a discriminated union (threshold-based XOR custom), making `shouldComplete` mutually exclusive with `minSuccessful`, `toleratedFailureCount`, and `toleratedFailurePercentage` at compile time; a runtime guard throws for plain-JavaScript callers.
- `shouldComplete` returns a `CompletionDecision` (`{ complete, outcome? }`) instead of a boolean. Added `completeBatch(outcome?)` and `continueBatch()` factory helpers.
- Split the `CUSTOM_COMPLETION` reason into `CUSTOM_COMPLETION_SUCCEEDED` and `CUSTOM_COMPLETION_FAILED`. The decision outcome is authoritative for the overall result: `CUSTOM_COMPLETION_FAILED` makes `BatchResult.status` FAILED and `throwIfError()` throw even when no individual item failed.
- Update unit tests, the parallel/should-complete example, and its history.


packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/concurrent-execution-handler.ts
```diff
@@ -59,20 +59,24 @@ export class ConcurrencyController<Logger extends DurableLogger> {
     // Check tolerance first, before checking if all completed
     const completion = config.completionConfig;
 
-    // A custom shouldComplete predicate takes full precedence over the
-    // threshold fields. All items finishing always wins, otherwise the
-    // predicate decides whether the batch was completed early.
+    // A custom shouldComplete predicate owns the completion decision. Its
+    // returned outcome determines whether the early completion is reported as
+    // a success or a failure. If the predicate does not complete the batch,
+    // the batch is ending because every item finished (ALL_COMPLETED).
```
yaythomas, line 104: PR mentions a runtime guard against null? 

e.g. throw when shouldComplete is present alongside any of minSuccessful/toleratedFailureCount/toleratedFailurePercentage

packages/aws-durable-execution-sdk-js/src/types/batch.ts
```diff
@@ -113,7 +118,49 @@ export interface BatchResult<TResult> {
...
+      complete: true;
+      /**
+       * Whether this completion is an overall success or failure.
+       *
+       * `FAILED` marks the whole batch as failed (its
+       * {@link BatchResult.status} is `FAILED` and {@link BatchResult.throwIfError}
+       * throws) even when no individual item failed — for example when a
+       * required quorum can no longer be met. Defaults to `SUCCEEDED`.
+       */
+      outcome?: BatchItemStatus.SUCCEEDED | BatchItemStatus.FAILED;
```
yaythomas: could argue this should be its own enum rather than borrowing items (i.e this enum is for the whole decision)

yaythomas (dismissed):

packages/aws-durable-execution-sdk-js/src/types/batch.ts
ParidelPooya: Addressed

ParidelPooya: Addressed

yaythomas (approved):

packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/concurrent-execution-handler.ts
ParidelPooya: Addressed

PR #715, opened 2026-07-14
feat(sdk): test connection button for Workflow Insight destination

ParidelPooya:
Bug bash feedback: switching the Explorer to S3 + Athena closes the Settings modal almost immediately, with no clear signal that the destination is configured correctly or that the backing resources exist. Adds a "Test connection" button to the Data Source tab that runs read-only checks against the current (unsaved) form values and reports a per-check pass/fail inline — the modal stays open — so users can confirm a destination is reachable and its config complete before saving or querying.

Host:
- New destinationTest.ts: testDestination(cfg) returns { ok, summary, checks[] } and dispatches per destination type. All probes are read-only.
    - S3 + Athena: required-field completeness; a note when neither workgroup nor result location is set (the "primary workgroup with no output location" trap); Glue table existence (reported as "will be created on Save" when absent, not a failure); and a real SELECT 1 that exercises the workgroup / output location / permissions.
    - DynamoDB: DescribeTable. Aurora: SELECT 1 via the RDS Data API. SQS: GetQueueAttributes. CloudWatch Logs: exact-name DescribeLogGroups.
- config.ts: factor the settings normalization out of readConfig into a shared helper and add configFromWireSettings(), so the test can validate the webview's unsaved values without persisting them.
- extension.ts: handle a new "testDestination" message; post the outcome back as a "destinationTestResult" payload (not the global error toast) so it renders inline in the modal.

Webview:
- SettingsModal.tsx: "Test connection" button plus a success/error Alert listing each check; the result clears when the modal reopens or the destination changes. App.tsx/types.ts: message plumbing and shared report types.

Tests: destinationTest.test.ts (per-destination success, connectivity failure, and completeness paths) and config.test.ts (configFromWireSettings mapping, defaults, and coercion). Host suite 156 passing; host + webview builds green.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight-vscode/src/config.test.ts
+++ b/packages/aws-durable-execution-sdk-js-insight-vscode/src/config.test.ts
@@ -1,0 +1,70 @@
+jest.mock("vscode", () => ({ workspace: { getConfiguration: jest.fn() } }), {
+  virtual: true,
+});
+jest.mock("@aws-sdk/credential-providers", () => ({
+  fromIni: jest.fn(),
+  fromNodeProviderChain: jest.fn(),
+}));
+
+import { configFromWireSettings } from "./config";
+
+describe("configFromWireSettings", () => {
+  it("applies defaults for an empty payload", () => {
+    const cfg = configFromWireSettings({});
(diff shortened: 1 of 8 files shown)


yaythomas (approved):
:shipit:

PR #716, opened 2026-07-14
feat(sdk): default to Claude Sonnet 5 and add a Bedrock model picker in Workflow Insight

ParidelPooya:
The previous default Bedrock model for the Workflow Insight extension's AI features (us.anthropic.claude-sonnet-4-20250514-v1:0) is deprecated. Updates the default to the Claude Sonnet 5 US inference profile, makes the model field much easier to get right, and adapts the Converse calls to Sonnet 5's inference parameters.

- Default model id is now "us.anthropic.claude-sonnet-5" (the US cross-region inference profile, matching the prior default's shape) in config.ts, DEFAULT_SETTINGS, the package.json setting default, the field placeholder, and the README example.
- The Bedrock Model ID field is now an Autosuggest: it offers the models your account can use as suggestions AND still accepts any value typed directly (useful for profiles/regions the list doesn't surface).
- New "List available models" button fetches models for the currently configured Region and AWS Profile via a new host module (bedrockModels.ts):
    - Combines system-defined inference profiles (ListInferenceProfiles, paginated) with on-demand, ACTIVE, text-output foundation models (ListFoundationModels), de-duplicated and sorted.
    - Base ids that only support the INFERENCE_PROFILE type are skipped (not directly invokable — the matching profile already covers them).
    - Reflects models AVAILABLE in the Region; a listed id may still require model access to be granted in the Bedrock console (noted in the UI).
- Adds the @aws-sdk/client-bedrock dependency for the control-plane list calls.
- Wiring: a "listModels" message resolves the region/profile from the unsaved form (configFromWireSettings) and posts a "bedrockModels" result back; failures render inline in the Settings modal rather than as a global toast.
- Temperature handling: Claude Sonnet 5 (adaptive reasoning always on) rejects the `temperature` inference parameter ("`temperature` is deprecated for this model"). A new bedrockConverse.ts wraps every Converse call: it drops temperature up front for models known to reject it (Claude 5-gen and newer) and, as a safety net, retries once without temperature on any error that mentions it. All three llm.ts call sites and the agentLoop.ts call route through this helper. The helper lives in its own vscode-free module so the agent loop doesn't pull the VS Code API into its import graph.

Tests: bedrockModels.test.ts (merge/dedupe/sort, filtering, pagination) and bedrockConverse.test.ts (model detection, up-front strip, error-retry, and non-temperature errors rethrown). Host suite 168 passing; host + webview typecheck and builds green.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight-vscode/src/bedrockConverse.test.ts
+++ b/packages/aws-durable-execution-sdk-js-insight-vscode/src/bedrockConverse.test.ts
@@ -1,0 +1,102 @@
+const send = jest.fn();
+jest.mock("@aws-sdk/client-bedrock-runtime", () => ({
+  BedrockRuntimeClient: jest.fn(() => ({ send })),
+  ConverseCommand: jest.fn((input) => ({ __t: "converse", input })),
+}));
+
+import { sendConverse, modelRejectsTemperature } from "./bedrockConverse";
+import { ConverseCommand } from "@aws-sdk/client-bedrock-runtime";
+
+const client = { send } as never;
+
+function lastInput() {
+  const calls = (ConverseCommand as unknown as jest.Mock).mock.calls;
+  return calls[calls.length - 1][0];
+}
+
+beforeEach(() => jest.clearAllMocks());
+
+describe("modelRejectsTemperature", () => {
+  it("flags Claude 5-gen (and newer) models", () => {
+    expect(modelRejectsTemperature("us.anthropic.claude-sonnet-5")).toBe(true);
+    expect(modelRejectsTemperature("anthropic.claude-opus-5")).toBe(true);
+    expect(modelRejectsTemperature("global.anthropic.claude-haiku-9")).toBe(
+      true,
+    );
+    // two-digit versions from 10 up
+    expect(modelRejectsTemperature("anthropic.claude-sonnet-10")).toBe(true);
+    expect(modelRejectsTemperature("anthropic.claude-opus-12")).toBe(true);
+    // Opus 4.7+ rejects temperature despite being a 4-generation id
+    expect(modelRejectsTemperature("us.anthropic.claude-opus-4-7")).toBe(true);
+    expect(modelRejectsTemperature("anthropic.claude-opus-4-9")).toBe(true);
+  });
+
+  it("does not flag older models that accept temperature", () => {
+    expect(
+      modelRejectsTemperature("us.anthropic.claude-sonnet-4-20250514-v1:0"),
+    ).toBe(false);
+    expect(
+      modelRejectsTemperature("anthropic.claude-3-5-sonnet-20241022-v2:0"),
+    ).toBe(false);
+    // Opus 4.5/4.6 still accept temperature (only 4.7+ rejects)
+    expect(modelRejectsTemperature("anthropic.claude-opus-4-5")).toBe(false);
+    // a hypothetical zero-padded low version must not false-match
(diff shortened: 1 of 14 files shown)


packages/aws-durable-execution-sdk-js-insight-vscode/src/extension.ts
```diff
@@ -376,6 +380,26 @@ class ExplorerPanel {
...
+    // list matches what the user is about to save. Errors are reported inline
+    // via the same payload rather than the global error toast.
+    const cfg = configFromWireSettings(settings);
+    try {
+      const models = await listBedrockModels({
+        region: cfg.region,
+        credentials: resolveCredentials(cfg.awsProfile),
+      });
+      this.post({ type: "bedrockModels", models });
+    } catch (err) {
```
yaythomas, line 394: wipes any previously fetched suggestion list in the webview (App.tsx unconditionally does setBedrockModels(msg.models)). If a user loaded a good list, then retried with a typo'd profile name, they lose the suggestions too. Consider omitting models on the error payload (or having the webview keep the old list when error is set).

packages/aws-durable-execution-sdk-js-insight-vscode/src/bedrockConverse.test.ts
```diff
@@ -0,0 +1,97 @@
...
+const client = { send } as never;
+
+function lastInput() {
+  const calls = (ConverseCommand as unknown as jest.Mock).mock.calls;
+  return calls[calls.length - 1][0];
+}
+
+beforeEach(() => jest.clearAllMocks());
+
+describe("modelRejectsTemperature", () => {
```
yaythomas, line 19: Opus 4.7 (us.anthropic.claude-opus-4-7) also rejects temperature on Bedrock (same "temperature is deprecated for this model" 400, there are several public issue reports). The regex only matches version 5+, so every Opus 4.7 call pays a guaranteed-failing first request before the retry net catches it. Maybe extend the known-reject set to cover the 4.7 generation, e.g. claude-opus-4-[7-9], or add it to the doc comment as a deliberate omission?

packages/aws-durable-execution-sdk-js-insight-vscode/src/bedrockModels.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (!m.modelId) continue;
+      const active = m.modelLifecycle?.status === "ACTIVE";
+      const text = (m.outputModalities ?? []).includes("TEXT");
+      const onDemand = (m.inferenceTypesSupported ?? []).includes("ON_DEMAND");
+      if (active && text && onDemand) ids.add(m.modelId);
+    }
+
+    return Array.from(ids).sort((a, b) => a.localeCompare(b));
+  } finally {
+    clearTimeout(timer);
```
yaythomas, line 83: destroy()?

packages/aws-durable-execution-sdk-js-insight-vscode/src/extension.ts
ParidelPooya: addressed

packages/aws-durable-execution-sdk-js-insight-vscode/src/bedrockModels.ts
ParidelPooya: addressed

packages/aws-durable-execution-sdk-js-insight-vscode/src/bedrockConverse.test.ts
ParidelPooya: addressed

yaythomas (approved):

PR #717, opened 2026-07-14
perf(sdk): cut checkpoint and batch hot-path CPU self-time by ~48%

ParidelPooya:
CPU profiling of the integration examples against the local test runner (with SDK
  source-level attribution) showed ~78% of SDK CPU self-time concentrated in three files:
  `checkpoint-manager.ts`, `serdes.ts`, and `batch-result.ts`. This PR removes the redundant
  work in those hot paths with **no wire-format or replay-semantic changes** — checkpointed
  payload shapes and event histories are byte-identical.

  **Changes:**

  - **`checkpoint-manager.ts`**
    - Compute each checkpoint item's wire size once at enqueue via `Buffer.byteLength`
      (previously `JSON.stringify` + `TextEncoder.encode` per item on *every* `processQueue`
      pass — for large payloads this re-serialized 100KB+ items repeatedly just to read a length)
    - Drain batches with a cursor + single `slice()` (previously O(n²) per-item `shift()`)
    - Resolve waiting operations via a hashed-id reverse index (previously MD5-hashed every
      tracked operation on each status change)
    - Hot-path log arguments are now lazy thunks
  - **`batch-result.ts`** — success/failure/started counts computed in a single pass at
    construction; `succeeded()`/`failed()`/`started()` memoized. Native `#private` fields keep
    `JSON.stringify` output identical to the previous getter-based implementation.
  - **`concurrent-execution-handler.ts`** — `Map`-based index join for `itemStatuses`
    (previously O(n²) `find`-inside-`map`)
  - **`step-id-utils.ts`** — `hashId` memoized (bounded cache, 10k entries)
  - **`logger.ts`** — `DURABLE_VERBOSE_MODE` cached at module load (reading `process.env` is a
    native call, and `log()` sits in checkpoint hot paths); `refreshLogConfig()` added for
    callers that mutate the env var after load. `run-durable.ts` refreshes the cache after its
    assignment, and the testing CLI now calls `setupEnvironment` *before* the handler import
    loads the SDK.
  - **`scripts/profile-sdk.ts`** (examples) — new profiling tool that runs examples against the
    copied SDK source (same mechanism as the coverage scripts) and emits Chrome
    DevTools-compatible `.cpuprofile` files with per-SDK-file attribution.

  ### Demo/Screenshots

  Profile comparison (identical parameters: 5 iterations, 100µs sampling, same example set):

  | SDK file | Baseline hits | Optimized hits | Δ |
  |----------|--------------:|---------------:|---:|
  | **Total SDK self-time** | **4,438** | **2,310** | **−47.9%** |
  | `utils/checkpoint/checkpoint-manager.ts` | 2,295 | 791 | −66% |
  | `utils/serdes/serdes.ts` | 1,111 | 851 | −23% |
  | `handlers/concurrent-execution-handler/batch-result.ts` | 581 | 448 | −23% |
  | `utils/logger/logger.ts` | 51 | 1 | −98% |
  | `utils/step-id-utils/step-id-utils.ts` | 39 | 9 | −77% |

  Reproduce with:
  `npx tsx scripts/profile-sdk.ts --cpu --examples comprehensive-operations,map/large-scale,parallel/basic,step/with-retry,promise/all --iterations 5`


packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/batch-result.ts
```diff
@@ -87,30 +87,63 @@ function reconstructBatchError(errorObject: SerializedBatchError): Error {
...
+    this.#failedMemo ??= this.all.filter(
       (item): item is BatchItem<R> & { error: ChildContextError } =>
         item.status === BatchItemStatus.FAILED && item.error !== undefined,
     );
+    return this.#failedMemo;
   }
 
   started(): Array<BatchItem<R> & { status: BatchItemStatus.STARTED }> {
-    return this.all.filter(
+    this.#startedMemo ??= this.all.filter(
```
yaythomas, line 145: these methods now return the same array instance on every call. Previously result.succeeded().sort(...) or .reverse() mutated a throwaway copy

now it mutates the cache and every later call observes it.

The new TSDoc on BatchResult.all covers mutating all, but not mutating the returned views. Maybe amend TSDoc annotation, or return Object.freeze(...)d arrays?

packages/aws-durable-execution-sdk-js-examples/package.json
```diff
@@ -39,6 +39,9 @@
     "generate-sam-template": "tsx scripts/generate-sam-template.ts",
     "test": "jest && vitest run",
     "test:integration": "NODE_ENV=integration jest --config jest.config.integration.js",
+    "profile": "npx tsx scripts/profile-sdk.ts --cpu",
```
yaythomas: nit: The new scripts use npx tsx while the sibling scripts (generate-sam-template) call tsx directly.

packages/aws-durable-execution-sdk-js-examples/scripts/profile-sdk.ts
```diff
@@ -0,0 +1,567 @@
...
+  for (let i = 0; i < args.length; i++) {
+    switch (args[i]) {
+      case "--cpu":
+        opts.cpu = true;
+        break;
+      case "--heap":
+        opts.heap = true;
+        break;
+      case "--iterations":
+        opts.iterations = parseInt(args[++i], 10);
```
yaythomas: unvalidated. a typo gives NaN and every example silently runs zero iterations.

packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/batch-result.ts
ParidelPooya: fixed

packages/aws-durable-execution-sdk-js-examples/scripts/profile-sdk.ts
ParidelPooya: fixed

yaythomas (approved):

PR #719, opened 2026-07-14
feat(sdk): Add executionStartTimestamp to InvocationBaseInfo

SilanHe:
*Issue #, if available:*

*Description of changes:* Add optional executionStartTimestamp field to InvocationBaseInfo so it is available on all onInvocation* hook parameters (InvocationInfo and InvocationEndInfo). The timestamp is derived from
initialExecutionEvent.StartTimestamp when available (should always be available).

Updated all test fixtures constructing InvocationInfo or InvocationEndInfo to include the new field.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
+++ b/packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
@@ -95,6 +95,7 @@ async function runHandler<
     executionArn: executionContext.durableExecutionArn,
     executionInput: customerHandlerEvent,
     operations: allOperations,
+    executionStartTimestamp: initialExecutionEvent?.StartTimestamp ?? undefined,
   };
 
   const invocationInfo: InvocationInfo = {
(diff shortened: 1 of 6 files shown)


packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
```diff
@@ -95,6 +95,9 @@ async function runHandler<
     executionArn: executionContext.durableExecutionArn,
     executionInput: customerHandlerEvent,
     operations: allOperations,
+    executionStartTimestamp: initialExecutionEvent?.StartTimestamp
+      ? new Date(initialExecutionEvent.StartTimestamp)
```
ParidelPooya: nit: StartTimestamp is already Date | undefined (the SDK deserializes it). Re-wrapping in new Date(...) is unnecessary and inconsistent with the existing pattern in operation.ts, which assigns the SDK's Date directly.

ParidelPooya (dismissed):
Untested happy path — the only test touching the real with-durable-execution.ts code path asserts executionStartTimestamp: undefined in both cases. The branch where a real timestamp gets set is never exercised end-to-end; other updated tests just hardcode the field on manually-built fixtures.

packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
SilanHe: ah I can remove the wrap

ParidelPooya (approved):

PR #721, opened 2026-07-15
fix(sdk): default Workflow Insight example auto-invoke schedule to off

ParidelPooya:
The example/demo CDK stack (packages/aws-durable-execution-sdk-js-insight/cdk) defaulted lambda.autoInvoke.enabled to true, which provisions a dispatcher Lambda + an EventBridge rule that re-invokes the example durable function every few minutes. That keeps generating billable Lambda invocations (and downstream destination writes) indefinitely until the stack is torn down — an easy way to incur cost without realizing it.

- Flip lambda.autoInvoke.enabled to false so auto-invocation is strictly opt-in. Deploying the example function itself stays on (it's free at rest); only the recurring scheduled invocation is now off by default.
- Document the cost rationale in the config so it's clear before enabling.

Also makes cdk/config.json self-documenting: JSON has no comments, so following the file's existing `_comment` convention, every field now has a sibling `<field>Comment` description (plus a per-section `_comment`), and the top-level `_comment` notes that auto-invoke is off by default. These keys are ignored by the CDK code, which reads config values by explicit key (no key enumeration), so behavior is unchanged and the stack still typechecks.

Adds a stack test asserting no EventBridge rule is synthesized with the default config. CDK typecheck + stack tests (11) pass.


yaythomas (approved):
:shipit:

PR #723, opened 2026-07-15
feat(sdk): add Amazon OpenSearch support to Workflow Insight extension

ParidelPooya:
Adds Amazon OpenSearch as a queryable destination in the Workflow Insight VS Code extension, alongside DynamoDB / Aurora / Redshift / S3-Athena / CloudWatch Logs. OpenSearch isn't SQL-native and Amazon OpenSearch Service requires SigV4-signed HTTPS, so this uses the OpenSearch SQL plugin for tabular queries.

- New src/opensearch.ts: runOpenSearchQuery (POST _plugins/_sql, normalizes {schema, datarows} to the shared result shape), fetchOpenSearchRecord (direct GET <index>/_doc/<executionArn> for the row drill-down), and pingOpenSearch (connectivity probe). All requests are SigV4-signed (service "es") with @smithy/signature-v4 + @aws-crypto/sha256-js over native fetch, using the extension's resolved credentials.
- schema.ts: new "opensearch" DestinationType and an OpenSearch-SQL prompt (record schema, dialect, few-shots). The index name is hyphenated so it is backtick-quoted; string fields are queried by plain name (the SQL plugin does NOT expose the .keyword multifield); the operations array-of-objects cannot be unnested via the SQL plugin.
- config.ts / package.json settings: opensearchEndpoint + opensearchIndex. Wired through extension.ts (query + record-fetch routing, index as table name) and destinationTest.ts (SigV4 ping). Settings webview updated (SettingsModal, types, starter query). Adds @smithy/signature-v4 + @aws-crypto/sha256-js deps.

Example CDK stack + exporter fixes (validated end-to-end against a live OpenSearch Service domain):

- OpenSearchExporter: send the SigV4-signed header set verbatim instead of merging a capital-cased "Content-Type" on top of the signer's lowercase "content-type". Sending both makes fetch collapse them case-insensitively into a value that no longer matches the canonical request SigV4 signed, so the domain rejected every index request with 403 Forbidden (silent, since the plugin runs exporters best-effort). Also include the response body in the thrown error so future failures are diagnosable.
- stack.ts: pass the domain endpoint to the example function (hoisted so the env var can be set) and output it, so the example function's OpenSearchExporter indexes records.

Tests: unit tests for runOpenSearchQuery/fetchOpenSearchRecord/pingOpenSearch (mocked signer + fetch: normalization, error surfacing, 404, _source flatten) and extended config/destinationTest coverage.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight-vscode/src/destinationTest.ts
+++ b/packages/aws-durable-execution-sdk-js-insight-vscode/src/destinationTest.ts
@@ -9,6 +9,7 @@ import { resolveCredentials } from "./config";
 import { tableExists, runAthenaQuery } from "./athena";
 import { runAuroraQuery } from "./aurora";
 import { runRedshiftQuery } from "./redshift";
+import { pingOpenSearch } from "./opensearch";
 
 /**
  * Result of a single check within a destination test (e.g. "Glue table exists",
@@ -86,6 +87,8 @@ export async function testDestination(
       return testAurora(cfg, credentials);
     case "redshift":
       return testRedshift(cfg, credentials);
+    case "opensearch":
+      return testOpenSearch(cfg, credentials);
     case "sqs":
       return testSqs(cfg, credentials);
     case "cloudwatch-logs-exporter":
@@ -267,6 +270,35 @@ async function testRedshift(
         sql: "SELECT 1",
       });
       return "Connected and ran a statement via the Redshift Data API.";
+    }),
+  );
+  return report(checks);
+}
+
+async function testOpenSearch(
+  cfg: InsightConfig,
+  credentials: ReturnType<typeof resolveCredentials>,
+): Promise<DestinationTestReport> {
+  const checks: DestinationCheck[] = [];
+  const missing: string[] = [];
+  if (!cfg.opensearchEndpoint) missing.push("Domain endpoint");
+  checks.push({
+    label: "Required fields",
+    ok: missing.length === 0,
+    detail:
+      missing.length === 0
+        ? `Endpoint "${cfg.opensearchEndpoint}", index "${cfg.opensearchIndex}".`
+        : `Missing: ${missing.join(", ")}.`,
+  });
+  if (missing.length > 0) return report(checks);
+
+  checks.push(
+    await probe("OpenSearch connection (SigV4)", async () => {
+      return await pingOpenSearch({
+        region: cfg.region,
+        credentials,
+        endpoint: cfg.opensearchEndpoint,
+      });
     }),
   );
   return report(checks);
--- a/packages/aws-durable-execution-sdk-js-insight-vscode/src/destinationTest.ts
+++ b/packages/aws-durable-execution-sdk-js-insight-vscode/src/destinationTest.ts
@@ -9,7 +9,7 @@ import { resolveCredentials } from "./config";
 import { tableExists, runAthenaQuery } from "./athena";
 import { runAuroraQuery } from "./aurora";
 import { runRedshiftQuery } from "./redshift";
-import { pingOpenSearch } from "./opensearch";
+import { pingOpenSearch, countOpenSearchDocs } from "./opensearch";
 
 /**
  * Result of a single check within a destination test (e.g. "Glue table exists",
@@ -299,6 +299,21 @@ async function testOpenSearch(
         credentials,
         endpoint: cfg.opensearchEndpoint,
       });
+    }),
+  );
+
+  // Second probe: GET <index>/_count. Catches a mistyped index name at test
+  // time (a 404 is a soft pass — the index is created on first export, so an
+  // empty/fresh domain shouldn't hard-fail).
+  checks.push(
+    await probe(`Index "${cfg.opensearchIndex}"`, async () => {
+      const n = await countOpenSearchDocs(
+        { region: cfg.region, credentials, endpoint: cfg.opensearchEndpoint },
+        cfg.opensearchIndex,
+      );
+      return n === undefined
+        ? `Index "${cfg.opensearchIndex}" not found yet — it's created on first export. Double-check the name if you expected data.`
+        : `Index "${cfg.opensearchIndex}" exists with ${n} document(s).`;
     }),
   );
   return report(checks);
--- a/packages/aws-durable-execution-sdk-js-insight-vscode/src/opensearch.ts
+++ b/packages/aws-durable-execution-sdk-js-insight-vscode/src/opensearch.ts
@@ -38,21 +38,35 @@ function cellToString(value: unknown): string {
   return String(value);
 }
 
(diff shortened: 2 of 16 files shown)


yaythomas (commented):
Really nice addition — the dialect prompt is clearly written from live validation (the `.keyword` non-resolution and array-of-objects limitations are exactly the sharp edges of the SQL plugin), and the OpenSearchExporter header fix is a great catch with a first-class explanatory comment. A few small comments inline, nothing blocking.

packages/aws-durable-execution-sdk-js-insight-vscode/src/schema.ts
```diff
@@ -10,6 +10,8 @@
  *   query operations by name via a JSONB/JSONPath predicate on the array.
  * - "redshift": Redshift SQL; record_json is a SUPER column — navigate/unnest it
  *   with PartiQL (dot/bracket paths, `FROM tbl t, t.record_json.operations o`).
+ * - "opensearch": Amazon OpenSearch; queried via the SQL plugin (_plugins/_sql).
+ *   String fields are text+keyword — filter/group on the `.keyword` subfield.
```
yaythomas: the header says to filter/group on `.keyword` but `RECORD_SCHEMA_OPENSEARCH` below (and the PR body) say the SQL plugin can't resolve `.keyword` — please could you align the header with the prompt?

packages/aws-durable-execution-sdk-js-insight-vscode/src/destinationTest.ts
```diff
@@ -272,6 +275,35 @@ async function testRedshift(
...
+    ok: missing.length === 0,
+    detail:
+      missing.length === 0
+        ? `Endpoint "${cfg.opensearchEndpoint}", index "${cfg.opensearchIndex}".`
+        : `Missing: ${missing.join(", ")}.`,
+  });
+  if (missing.length > 0) return report(checks);
+
+  checks.push(
+    await probe("OpenSearch connection (SigV4)", async () => {
```
yaythomas, line 296: suggestion: a second probe (signed `GET /<index>/_count`) would catch a typo'd index name at test time, mirroring the depth Athena gets from `tableExists` — optional, the Redshift precedent is connection-only.

packages/aws-durable-execution-sdk-js-insight-vscode/src/opensearch.ts
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Run a query via the OpenSearch SQL plugin (POST _plugins/_sql) and normalize
+ * the {schema, datarows} response into the shared tabular result shape.
+ */
+export async function runOpenSearchQuery(
+  opts: OpenSearchConnection & { sql: string },
+): Promise<OpenSearchQueryResult> {
+  const endpoint = opts.endpoint.replace(/\/$/, "");
+  const url = `${endpoint}/_plugins/_sql`;
```
yaythomas: question: does `plugins.query.size_limit` (default 200) silently cap a verbatim query with no explicit LIMIT? if so, maybe worth a note in the endpoint setting description — I could easily be mistracing this.

packages/aws-durable-execution-sdk-js-insight-vscode/src/opensearch.ts
```diff
@@ -0,0 +1,190 @@
...
+}
+
+/**
+ * Run a query via the OpenSearch SQL plugin (POST _plugins/_sql) and normalize
+ * the {schema, datarows} response into the shared tabular result shape.
+ */
+export async function runOpenSearchQuery(
+  opts: OpenSearchConnection & { sql: string },
+): Promise<OpenSearchQueryResult> {
+  const endpoint = opts.endpoint.replace(/\/$/, "");
```
yaythomas: nit: `endpoint.replace(/\/$/, "")` is repeated in all three exported functions — could live once in `signedFetch`.

packages/aws-durable-execution-sdk-js-insight-vscode/src/opensearch.ts
```diff
@@ -0,0 +1,190 @@
...
+    "POST",
+    url,
+    JSON.stringify({ query: opts.sql }),
+  );
+
+  const text = await res.text();
+  if (!res.ok) {
+    // The SQL plugin returns a JSON error body with reason/details — surface it
+    // so the model (agent mode) can correct the query.
+    let detail = text.slice(0, 800);
```
yaythomas: nit: error-detail truncation lengths vary (800/500/300 chars) across the three functions — trivial to harmonize.

packages/aws-durable-execution-sdk-js-insight-vscode/src/opensearch.ts
```diff
@@ -0,0 +1,190 @@
...
+  return String(value);
+}
+
+/**
+ * SigV4-sign and send a request to the OpenSearch domain. Amazon OpenSearch
+ * Service authorizes with SigV4 over the "es" service; the caller's resolved
+ * credentials (profile/instance role) must be allow-listed in the domain's
+ * access policy. Uses native fetch — no OpenSearch client library.
+ *
+ * No query-string params are used (paths only), so the signed canonical
```
yaythomas: no-action note: the "no query-string params" invariant is only by convention — `if (u.search) throw` would make a future `?pretty` 403 self-diagnosing.

packages/aws-durable-execution-sdk-js-insight-vscode/src/schema.ts
ParidelPooya: Fixed

packages/aws-durable-execution-sdk-js-insight-vscode/src/opensearch.ts
ParidelPooya: Addressed

packages/aws-durable-execution-sdk-js-insight-vscode/src/opensearch.ts
ParidelPooya: Addressed

packages/aws-durable-execution-sdk-js-insight-vscode/src/opensearch.ts
ParidelPooya: Addressed

packages/aws-durable-execution-sdk-js-insight-vscode/src/opensearch.ts
ParidelPooya: Addressed

packages/aws-durable-execution-sdk-js-insight-vscode/src/destinationTest.ts
ParidelPooya: Addressed

yaythomas (approved):

PR #728, opened 2026-07-16
feat(sdk): expose retry attempt number on waitForCondition context

ParidelPooya:
Add `attempt` field to `WaitForConditionContext` so check functions can read the current retry attempt, mirroring the `StepContext.attempt` field added for step() in #707 (and matching the Java SDK, where waitForCondition's check function already reuses StepContext.getAttempt()).

Previously only the wait strategy's positional `attempt` argument had this information; the check function itself had no way to know which attempt it was on without tracking state manually.

- Convert WaitForConditionContext from a type alias to an interface extending OperationContext, adding attempt: number (1-based).
- Populate it in the wait-for-condition handler from the existing currentAttempt already used for waitStrategy and logging.
- Update the waitForCondition() and WaitForConditionCheckFunc TSDoc.
- Add unit tests covering the first-attempt and post-retry values.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.test.ts
+++ b/packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.test.ts
@@ -357,6 +357,66 @@ describe("WaitForCondition Handler", () => {
       expect(waitStrategy).toHaveBeenCalledWith("result", 1);
     });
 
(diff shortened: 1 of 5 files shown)


yaythomas (approved):
api-extractor/durable-execution-sdk-js.api.json

#707 updated the api-extractor report for the same type → interface conversion on 
StepContext (canonicalReference ~StepContext:type → :interface, plus the docComment changes).

For follow-up, ~WaitForConditionContext:type and the waitForCondition()/WaitForConditionCheckFunc docComments stale in the report.

PR #741, opened 2026-07-21
fix(sdk): switch release workflow to draft-first, workflow_dispatch

ParidelPooya:
The workflow_insight-vscode-v0.1.0-alpha.0 and v0.1.0-alpha.1 release attempts both failed at the final 'gh release upload' step:

  HTTP 422: Cannot upload assets to an immutable release.

This repo has GitHub's immutable-releases protection enabled: once a release is PUBLISHED, its assets can never be added, changed, or removed. The previous workflow triggered on 'release: published' — by definition, after the release was already published and therefore already immutable. There was no way to attach assets afterward.

GitHub's own docs recommend the fix: create the release as a DRAFT, attach every asset while still a draft (drafts are exempt from immutability), then publish last. But 'release' events are explicitly not sent for created/edited/deleted activity on draft releases (per GitHub's events-that-trigger-workflows docs), so no release-event trigger can drive this sequence — only workflow_dispatch (run manually, before any release or tag exists) can.

Redesigned the workflow:
- Trigger: workflow_dispatch with a required 'version' input, validated in check-version against packages/aws-durable-execution-sdk-js-insight-vscode/ package.json's actual version before anything else runs.
- New create-draft-release job: creates the release as a draft (or reuses an existing draft for the same tag, so a partial-failure re-run doesn't duplicate it); refuses to touch a release that's already published.
- build-and-attach: unchanged internally, now uploads to the draft.
- verify-and-publish (renamed from verify-all-platforms): still uses '!cancelled()' to run even if a matrix leg failed, but now only publishes the release (gh release edit --draft=false) if all 4 .vsix files are confirmed present. If anything's missing, the release is left as an unpublished, non-immutable draft — re-running the workflow after a fix needs no manual cleanup.

Verified against the current  CLI manual: --draft on release create, --draft=false on release edit, and the isDraft JSON field on release view are all documented and match this design (gh's own docs describe exactly this create-draft/upload-assets/publish-last pattern for immutable releases).

### DIFF
--- a/.github/workflows/vscode-extension-release.yml
+++ b/.github/workflows/vscode-extension-release.yml
@@ -1,7 +1,8 @@
 # Builds the Workflow Insight Explorer VS Code extension for each supported
-# platform/arch and attaches the packaged .vsix files to the GitHub Release as
-# assets. This extension is not published to npm (it's "private": true) or,
-# during preview, to the VS Code Marketplace — see
+# platform/arch, creates a DRAFT GitHub Release, attaches all 4 platform
+# .vsix files to it, then publishes the release as the final step. This
+# extension is not published to npm (it's "private": true) or, during
+# preview, to the VS Code Marketplace — see
 # packages/aws-durable-execution-sdk-js-insight-vscode/README.md.
 # Testers/customers install it by downloading the .vsix matching their
 # platform from the release and side-loading it with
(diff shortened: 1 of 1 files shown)


.github/workflows/vscode-extension-release.yml
```diff
@@ -13,48 +14,83 @@
...
+    permissions:
+      contents: write # required to create the draft release
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Create draft release (or reuse an existing one for this tag)
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          TAG="${{ needs.check-version.outputs.tag_name }}"
```
yaythomas: could we pass the ${{ }} values into the scripts via env: instead of inline interpolation? e.g.

.github/workflows/vscode-extension-release.yml
```diff
@@ -152,42 +186,39 @@ jobs:
...
     # verification would never run in exactly the scenario it exists to
-    # catch (a partial release with some platforms missing). We still gate
-    # on check-version's own result explicitly, so this doesn't run at all
-    # for a release this workflow was never supposed to act on.
-    if: |
-      !cancelled() &&
-      needs.check-version.result == 'success' &&
-      needs.check-version.outputs.should_release == 'true'
+    # catch (a partial release with some platforms missing).
+    if: ${{ !cancelled() }}
```
yaythomas: `!cancelled() && needs.check-version.result == 'success' && ...` to bare `if: ${{ !cancelled() }}`.

 With !cancelled(), the job now runs even when check-version itself failed (e.g. version mismatch) — TAG expands to empty, gh release view "" errors, and the job fails red with a confusing "release not found" instead of being skipped. Not a safety issue (nothing gets published; set -euo pipefail catches it), but it turns every aborted run into two red jobs. Suggest if: ${{ !cancelled() && needs.check-version.result == 'success' }} — keeps the fail-fast-false rationale for build-and-attach legs while skipping cleanly when the version gate aborts.

.github/workflows/vscode-extension-release.yml
ParidelPooya: Addressed

.github/workflows/vscode-extension-release.yml
ParidelPooya: Fixed

yaythomas (approved):

PR #759, opened 2026-07-24
fix(sdk): stop summarized map/parallel replay from hanging

### LINKED ISSUE #751: Summarized map/parallel results replay non-deterministically; a custom summaryGenerator can hang replay
## Summary

When a map/parallel aggregate result exceeds the 256KB checkpoint limit, the SDK checkpoints `summaryGenerator(result)` instead of the full result and sets `ReplayChildren` (`run-in-child-context-handler.ts`, `executeChildContext`). On replay, the SDK reconstructs the `BatchResult` from child checkpoints plus that summary. The reconstruction reads exactly one field from the summary (`totalCount`) and re-derives everything else. Two consequences, both reproduced deterministically with `LocalDurableTestRunner` against the published packages:

1. **Default configuration**: after suspend/resume, the same line of user code observes a different `BatchResult` than the live run observed.
2. **Custom `summaryGenerator`**: any generator output that is not JSON with a numeric `totalCount` key makes replay fall back to live concurrent execution inside `ReplaySucceededContext` mode, where the batch can never complete. The replay invocation hangs and the execution makes no further progress.

Both are silent. No warning or error fires on any path.

## Repro 1: default generator, divergent replay (sdk 2.1.0 and 2.2.0)

```js
const handler = withDurableExecution(async (_event, context) => {
  const result = await context.map(
    "work",
    [0, 1, 2, 3, 4],
    async (ctx, _item, index) => {
      if (index >= 2) await ctx.wait(`item-wait-${index}`, { seconds: 300 });
      return "x".repeat(150 * 1024); // 2 successes > 256KB threshold
    },
    { completionConfig: { minSuccessful: 2 } },
  );
  const snapshot = {
    totalCount: result.totalCount,
    successCount: result.successCount,
    startedCount: result.startedCount,
    completionReason: result.completionReason,
    itemIndexes: result.all.map((i) => i.index),
  };
  observed.push(snapshot);                         // per-invocation observation
  await context.wait("after-map", { seconds: 1 }); // suspend, then replay
  return snapshot;
});
```

Observed (`LocalDurableTestRunner`, real timers):

```
live run observed: {"totalCount":5,"successCount":2,"startedCount":3,"completionReason":"MIN_SUCCESSFUL_REACHED","itemIndexes":[0,1,2,3,4]}
replay observed:   {"totalCount":2,"successCount":2,"startedCount":0,"completionReason":"MIN_SUCCESSFUL_REACHED","itemIndexes":[0,1]}
```

## Repro 2: custom generator, replay hangs (sdk 2.2.0, generator configurable since #696)

Same shape, plus:

```js
{
  maxConcurrency: 2,
  completionConfig: { minSuccessful: 2 },
  summaryGenerator: (r) => `processed ${r.successCount}/${r.totalCount} items`,
}
```

The live run completes the map (items 0 and 2 succeed, item 1 in flight) and suspends. The replay invocation never finishes. The identical workload with the default generator completes in about 2 seconds.

## Mechanism (all from source, current main)

- `concurrent-execution-handler.ts`, `executeItems`: on replay, only `totalCount` is read from the parsed summary. The default generators also write `successCount`, `failureCount`, `completionReason`, and `status`, and none of those are ever read back.
- `replayItems` reconstructs terminal children only, so live STARTED items are dropped from the replayed result. Its stop gate compares the replayed completed count against `totalCount`, but the live `totalCount` includes STARTED items, so the gate cannot fire after an early completion.
- `completionReason` is re-inferred from the reconstructed items via `getCompletionReason`. A custom `shouldComplete` predicate gets re-invoked on replay with the reconstructed (terminal-only) statuses.
- If the summary is missing, unparseable, or lacks `totalCount`, `executeItems` falls back to `executeItemsConcurrently` while the context is in `ReplaySucceededContext` mode. In that mode, a child whose checkpoint is non-terminal receives a never-resolving promise (`durable-context.ts`, `checkForNonResolvingPromise`). That gate fires before the pending step id is consumed, so every later item in the batch blocks on the same pending step id, including items with SUCCEEDED checkpoints. The batch promise can never settle.
- `totalCount` is used without runtime validation (`as number` cast). `{"totalCount": null}` passes both guards, and `0 >= null` is `true`, so replay returns an empty `BatchResult` for a batch that had live results.

## Impact

- Breaks the replay determinism contract: code after a summarized map observes different `totalCount` / `startedCount` / item lists across invocations.
- Contradicts the documented contract. The developer guide (map and parallel pages, TypeScript tab) states: "the checkpoint stores a compact JSON summary, which is for observability only." Replay functionally depends on that summary, and a customer who follows the docs and writes a free-form observability string walks directly into the hang in Repro 2.
- The real contract is undocumented and load-bearing: the type is `(result) => string`, but replay correctness silently requires JSON with a numeric `totalCount` whose meaning matches the SDK's internal use.
- Both failure modes are silent.

### Non-breaking (read side)

- Read `completionReason` from the summary when present instead of re-inferring it. This also stops re-invoking a custom `shouldComplete` predicate during replay.
- Use `successCount + failureCount` as the reconstruction target instead of `totalCount`. Both fields are already written by the default generators, so checkpoints from existing executions benefit on their next resume with no write-side change.
- Reconstruct STARTED placeholders for indexes up to the recorded `totalCount`. Items start in strict index order (`tryStartNext`), so the live started set is always the index prefix. Together with the two points above this reproduces the live aggregate shape (`totalCount`, counts, reason) from fields the default generators already write. Per-item statuses remain best-effort in one window: a child that checkpoints terminal between the completion decision and suspension is indistinguishable from one that completed before it, so exact per-item fidelity needs the recorded started set from the write-side change below.
- Validate the record: `totalCount` must be a non-negative integer, anything else is treated as missing.
- On a missing or malformed record, derive from child checkpoints instead of falling back to concurrent execution in `ReplaySucceededContext` mode. This fixes the hang, and a hang is not behavior anyone can be depending on.

### Write side: two options for making custom generators replay-safe

**Option 1: SDK envelope.** Wrap every generator's output verbatim under a `summary` key inside an SDK-owned record:

```json
{
  "totalCount": 5,
  "completionReason": "MIN_SUCCESSFUL_REACHED",
  "startedIndexes": [2, 3, 4],
  "summary": "<generator output, verbatim>"
}
```

What breaks: the stored payload shape visible in execution history. Softened to effectively non-breaking by the documented contract: the docs call this payload "for observability only," so no programmatic consumer of its shape was ever supported. A changelog callout should suffice, and the envelope makes the documented contract true, with customer output staying observability-only under `summary`.

**Option 2: mandatory fields, no envelope.** Require generator output to be JSON with the record fields, typed as an object instead of an opaque string:

```ts
summaryGenerator?: (result: BatchResult<TResult>) => {
  totalCount: number;
  completionReason: CompletionReason;
  startedIndexes: number[]; // items in flight at the completion decision
  [key: string]: unknown;   // free-form customer fields
};
```

What breaks: the TS signature, for every existing custom generator of an API that shipped two weeks ago (2.2.0, #696). Default generator payloads are unchanged, and old or invalid checkpoints fall back to checkpoint derivation. Unlike Option 1, this reverses the documented observability-only contract, and validation can check field types but not that customer-computed values (especially `startedIndexes`) are actually correct.

ParidelPooya:
When a `map`/`parallel` aggregate result exceeds the 256KB checkpoint limit, the
   SDK checkpoints a compact summary and reconstructs the `BatchResult` from child
   checkpoints on replay (ReplayChildren mode). Two failures came out of this path,
   both from issue #751:

   1. **Replay hang.** When the batch completed early (e.g. `minSuccessful`) with an
      item still in flight *between* two completed items, replaying the summarized
      map hung and the execution made no further progress.
   2. **Free-form custom summary hang.** A custom `summaryGenerator` whose output was
      not JSON with a numeric `totalCount` made replay fall back to concurrent
      execution in `ReplaySucceededContext` mode and hang there too.

   Fixes #751.

   ## Root cause (the hang)

   `replayItems` re-drives `runInChildContext` on the map/parallel **child context**,
   but `skipNextOperation` was bound to the **outer context** that owns the `map`
   call. When an in-flight item was skipped, the child context's step cursor was not
   advanced, so the next terminal item's `runInChildContext` peeked the pending
   non-terminal step and `checkForNonResolvingPromise` returned a never-resolving
   promise.

   This path only runs in the summarized (ReplayChildren) case, which no existing
   test covered — and prior unit tests mocked `runInChildContext`, hiding the gate.
   The bug was found and reproduced end-to-end with `LocalDurableTestRunner` via a
   new example (reproduction-first).

   ## Changes

   **Read side (`concurrent-execution-handler`)**
   - Skip on the same context `runInChildContext` runs on (`parentContext`) via
     `skipReplayStep`, so the non-resolving-promise gate no longer fires. Removed the
     misbound `skipNextOperation` parameter from the controller/factory.
   - Read `completionReason` from the summary instead of re-inferring it — this also
     stops re-invoking a custom `shouldComplete` predicate on replay. The recorded
     value is validated against the known `CompletionReason` set; an unrecognized or
     malformed value is ignored and the reason is re-inferred (so a bogus string can
     never leak into `BatchResult.completionReason`).
   - Always reconstruct from child checkpoints in `ReplaySucceededContext` mode rather
     than falling back to concurrent execution. Summarized replay returns the
     **completed** items; in-flight items are not rebuilt (the recorded
     `completionReason` carries the batch outcome).

   **Write side**
   - `composeSummaryGenerator`: always record the SDK's load-bearing metadata
     (`totalCount`/`successCount`/`failureCount`/`completionReason`/`status`) and store
     any custom generator's output verbatim under a `summary` key (observability-only).
     This makes a free-form custom summary parseable so replay takes the reconstruction
     path instead of the concurrent fallback.
   - Public `summaryGenerator?: (result) => string` signature is unchanged.

   ## Behavior note (scope)

   This PR targets the hang. For a batch that **completed early with in-flight items**
   and whose result was large enough to be summarized, replay reconstructs the
   completed items only — it does not rebuild the in-flight (`STARTED`) entries, so
   `startedCount` / the in-flight index list are not reproduced on that path. The
   non-summarized (small-result) path is unchanged and still replays the full cached
   result. This trade-off is intentional: the batch outcome is driven by the recorded
   `completionReason`, and code depending on the exact in-flight set across a
   summarized replay is not a supported pattern.

   ## Testing

   - New `map/custom-summary-generator-replay` example + integ test reproducing the
     hang end-to-end (small committed history fixture).
   - Updated `parallel/custom-summary-generator` test for the new envelope shape (custom
     output now nested under `summary`).
   - Updated `concurrent-execution-handler` unit/replay/two-phase tests for the new
     controller signature and completed-items-only reconstruction; added composer tests
     and a test that a bogus recorded `completionReason` is ignored and re-inferred.

   All green: SDK unit 1065 · examples integ 186 (+1 cloud-only skip) · examples vitest 2
   · `-testing` 1167 · `-otel` 160.

   ## Compatibility

   - The `summaryGenerator` type and the default checkpoint shape are unchanged.
   - For a **custom** `summaryGenerator` on `map`/`parallel`, the checkpointed summary is
     now an SDK envelope with the customer's output under `summary`, instead of the raw
     customer string. The developer guide already documents this payload as "for
     observability only," so no supported programmatic consumer is affected — worth a
     changelog callout.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/concurrent-execution-handler.ts
+++ b/packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/concurrent-execution-handler.ts
@@ -24,10 +24,29 @@ import {
   restoreBatchResult,
   createBatchResultSerdes,
 } from "./batch-result";
-import { defaultSerdes, AnySerdes } from "../../utils/serdes/serdes";
+import { AnySerdes } from "../../utils/serdes/serdes";
 import { ChildContextError } from "../../errors/durable-error/durable-error";
 import { TerminationManager } from "../../termination-manager/termination-manager";
 import { TerminationReason } from "../../termination-manager/types";
+
+/**
+ * Valid {@link CompletionReason} values, used to validate a reason read back
+ * from a checkpointed summary. Typed as `Record<CompletionReason, true>` so it
+ * is exhaustive by construction: adding a value to the {@link CompletionReason}
+ * union turns this into a compile error until the new value is listed here — the
+ * validator can never silently omit a reason (and no extra public API is
+ * exported to keep it in sync). A reason not present here (a malformed or
+ * hand-authored summary) is ignored so an arbitrary string can never leak into
+ * {@link BatchResult.completionReason}; replay re-infers the reason from the
+ * reconstructed items instead.
+ */
+const VALID_COMPLETION_REASONS: Record<CompletionReason, true> = {
+  ALL_COMPLETED: true,
+  MIN_SUCCESSFUL_REACHED: true,
+  FAILURE_TOLERANCE_EXCEEDED: true,
+  CUSTOM_COMPLETION_SUCCEEDED: true,
+  CUSTOM_COMPLETION_FAILED: true,
+};
 
 /**
  * Validates that a custom `shouldComplete` predicate is not combined with the
@@ -67,24 +86,28 @@ function validateCompletionConfig(
 export class ConcurrencyController<Logger extends DurableLogger> {
   constructor(
     private readonly operationName: string,
-    private readonly skipNextOperation: () => void,
 
     private readonly getDefaultSerdes?: () => AnySerdes,
   ) {}
 
-  private isChildEntityCompleted(
-    executionContext: ExecutionContext,
-    parentEntityId: string,
-    completedCount: number,
-  ): boolean {
-    const childEntityId = `${parentEntityId}-${completedCount + 1}`;
-    const childStepData = executionContext.getStepData(childEntityId);
-
-    return !!(
-      childStepData &&
-      (childStepData.Status === OperationStatus.SUCCEEDED ||
-        childStepData.Status === OperationStatus.FAILED)
-    );
+  /**
+   * Advance the step cursor of the context that {@link replayItems} re-drives
+   * child operations on, for a non-terminal item that is NOT re-executed on
+   * replay (one that was in flight at suspension, or never started).
+   *
+   * This MUST advance the SAME context that runInChildContext runs on
+   * (`parentContext` — the map/parallel child context), not the outer context
+   * that owns the map/parallel call. If a skipped in-flight item does not
+   * advance this cursor, the next terminal item's runInChildContext peeks the
+   * pending non-terminal step, `checkForNonResolvingPromise` returns a
+   * never-resolving promise, and replay hangs (issue #751). `skipNextOperation`
+   * is internal to the durable context, so it is reached through a narrow cast
+   * (the same pattern used to read `_stepPrefix`/`durableExecutionMode`).
+   */
+  private skipReplayStep(parentContext: DurableContext<Logger>): void {
+    (
+      parentContext as unknown as { skipNextOperation: () => void }
+    ).skipNextOperation();
   }
 
   private getCompletionReason<T, R>(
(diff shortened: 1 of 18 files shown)


packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/concurrent-execution-handler.ts
```diff
@@ -230,140 +231,193 @@ export class ConcurrencyController<Logger extends DurableLogger> {
...
-
-      // Calculate the child entity ID that runInChildContext will create
-      // It uses the parent's next step ID, which is parentEntityId-{counter}
       const childEntityId = `${parentEntityId}-${stepCounter + 1}`;
+      const childStepData = executionContext.getStepData(childEntityId);
+      const isTerminal =
+        !!childStepData &&
+        (childStepData.Status === OperationStatus.SUCCEEDED ||
+          childStepData.Status === OperationStatus.FAILED);
+      const wasStarted = !!childStepData;
```
yaythomas: f a map item finishes in the tiny gap between the batch deciding "done" (via minSuccessful) and the SDK sealing the checkpoint, its SUCCEEDED checkpoint lands anyway, so replay rebuilds that item as SUCCEEDED while the live run returned it as STARTED, giving your code different counts before and after a suspend.

as possible way around this is to record the started/completed index set in composeSummaryGenerator, and check it here?

packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/concurrent-execution-handler.ts
```diff
@@ -230,140 +231,193 @@ export class ConcurrencyController<Logger extends DurableLogger> {
...
+  private parseRecordedCompletionReason(
+    summaryPayload: unknown,
+  ): CompletionReason | undefined {
+    if (typeof summaryPayload !== "string") {
+      return undefined;
+    }
+    try {
+      const parsed = JSON.parse(summaryPayload);
+      if (parsed && typeof parsed === "object") {
+        const reason = (parsed as Record<string, unknown>).completionReason;
```
yaythomas, line 273: No validation? will `completionReason: "whatever"` end up in BatchResult.completionReason?

packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/concurrent-execution-handler.ts
ParidelPooya: Yes, but as dicussed, I removed this from PR

packages/aws-durable-execution-sdk-js/src/handlers/concurrent-execution-handler/concurrent-execution-handler.ts
ParidelPooya: Fixed

yaythomas (approved):

PR #767, opened 2026-07-28
fix(sdk): use safeDeserialize to fail on serdes errors

### LINKED ISSUE #754: [Bug]: waitForCondition silently resets to initialState when checkpointed state fails to deserialize
### Expected Behavior

When `waitForCondition` resumes and the checkpointed polling state cannot be deserialized, the operation fails with the serdes error.

The Java SDK behaves this way: `SerializableDurableOperation.deserializeResult` logs a warning and rethrows `SerDesException`, failing the operation.

### Actual Behavior

`executeCheckLogic` in `wait-for-condition-handler.ts` catches the deserialize failure with a bare `catch` and restarts the condition loop from `config.initialState`:

https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.ts#L183-L195

The failure is not logged. All polling progress is discarded and the check function re-runs from scratch. For a condition that accumulates state or counts toward a threshold this produces silently wrong results, and polling can extend past the configured attempt budget because the state driving the wait strategy restarts from zero.

Note the same handler already routes the post-check round-trip through `safeDeserialize` (which fails the execution via the termination manager). The resume-read site is the only spot that swallows instead.

Trigger conditions: a custom serdes whose `deserialize` throws, or a serdes/config change deployed while an execution is mid-poll.

### Steps to Reproduce

1. Configure `waitForCondition` with a custom serdes whose `deserialize` throws.
2. Let the operation checkpoint one continue decision and suspend.
3. Resume. The next invocation continues from `initialState` with no log line instead of failing.

### SDK Version

current main

### Is this a regression?

No

### Additional Context

Twin of aws/aws-durable-execution-sdk-python#574, found while auditing the wait_for_condition fix in aws/aws-durable-execution-sdk-python#541. Python has the same fallback (with logging). Java propagates.

Proposed fix: use the existing `safeDeserialize` path (or rethrow) so the serdes error flows through the operation's failure handling, matching Java. Behavior change, so v2.

jibin7jose:
Fixes #754
### Description
Replaces the bare `catch` and native `serdes.deserialize` call inside `wait-for-condition-handler.ts` with the existing `safeDeserialize` utility. 
Previously, if the checkpointed polling state could not be deserialized, the exception was silently swallowed and the condition loop silently restarted from `config.initialState`. 
Using `safeDeserialize` matches the behavior of the Java SDK and the recent Python SDK updates by properly logging the failure, rethrowing the `SerDesException`, and terminating the execution via the termination manager.


ParidelPooya:
Thanks for the contribution! Could you please add corresponding unit and integration tests to cover these changes?

jibin7jose:
Thanks for the review! I've added the requested unit and integration tests in the latest commit. Please let me know if there's anything else you'd like me to update.

ayushiahjolia:
The core fix looks good, thank you!

A few things to address:
1. Build break: SerDesException does not exist
The conformance test handler imports SerDesException from @aws/durable-execution-sdk-js, but this class isn't defined or exported anywhere in the SDK. This breaks the build - see [#785](https://github.com/aws/aws-durable-execution-sdk-js/pull/785). 

2. Conformance test requires a matching spec in the conformance repo
Adding a handler with TestDescription: ["6-14"] requires a corresponding 6-14.yaml requirement spec in aws/aws-durable-execution-conformance-tests. Without it the handler deploys but never gets validated. For a change this small, you could rely on this repo's own integration tests instead (see packages/aws-durable-execution-sdk-js-examples/).

3. Unit test only covers STARTED status
The fix applies to both OperationStatus.STARTED and OperationStatus.READY, but the test only exercises STARTED. Adding a parallel case for READY would fully cover the changed code path.

Please rebase your change based off main as well :)

ayushiahjolia:
Let us know when it's ready to review again!

I have also updated contributing guideline for guidance on Conformance tests - https://github.com/aws/aws-durable-execution-sdk-js/pull/791

jibin7jose:
I have addressed the review comments, migrated the test to examples, and fixed the SerDesException import issue. Note that the 10 conformance test checks are still failing, but this is because GitHub Actions doesn't provide AWS credentials to external forks. The code itself is ready for review

yaythomas:
Thanks for the update @jibin7jose! 

The core fix is right, and it's the part that matters. :-) 

On the new test: I ran it against both handler implementations, and it passes with *and* without the fix, so it can't detect the regression. The custom serdes throws on every `deserialize`, so it fails at the pre-existing post-check
round-trip on invocation 1, before any checkpoint or resume, and never reaches the STARTED/READY restore branch this PR changes.

It's not really easily fixable from inside a serdes: the checkpointed payload is the same string the post-check
already read, with the same `SerdesContext`, so anything that throws at restore has already thrown a invocation earlier. 

So I've pushed a commit to your branch that 
- removes that example,
- reverts `template_wait_for_condition.yaml` (its 251 changed lines were LF→CRLF only),
- and replaces two unused `handlerPromise` variables with the `void` call idiom used in
`concurrent-execution-handler.test.ts`.
- I also merged current `main` in.


I kept your two STARTED/READY unit tests, those *do* pin the branch: they fail against the old `try/catch` and pass with `safeDeserialize`. Thank you so much!

Since forks can't run them in CI, I ran the cloud checks locally: all ten conformance suites (171 tests, 0 failures) and the full integration suite on 22.x and 24.x. Green apart from two OpenTelemetry X-Ray tests that fail identically on clean `main`.

And yes, you're right, the conformance checks here fail at credential setup because forks don't get `AWS_REGION`, not because of your diff.

Thanks again for sticking with this!

I'm just waiting for the push I did to your branch to pass then I'll merge.

yaythomas (dismissed):

packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.ts
```diff
@@ -181,18 +181,14 @@ export const createWaitForConditionHandler = <Logger extends DurableLogger>(
...
-                durableExecutionArn: context.durableExecutionArn,
-              };
-              currentState = await serdes.deserialize(
-                checkpointData,
-                serdesContext,
-              );
-            } catch {
-              currentState = config.initialState;
-            }
+            currentState = await safeDeserialize(
```
ParidelPooya, line 184: Confirmed this is the right change, and confirmed it's covered: reverting just these lines to the old `try/catch` makes both new tests fail, so the branch is genuinely pinned.

Worth noting for the record that this removes the last direct `serdes.deserialize` call in `src/` — everything now goes through `safeDeserialize`, and this restore path now matches the `SUCCEEDED` replay path a few dozen lines above.

packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.test.ts
```diff
@@ -565,4 +566,119 @@ describe("WaitForCondition Handler", () => {
...
+
+      // Force safeDeserialize to fail for state deserialization
+      mockSafeDeserialize.mockImplementationOnce(async () => {
+        mockContext.terminationManager.terminate({
+          reason: TerminationReason.SERDES_FAILED,
+          message: "Failed to deserialize operation payload",
+          error: expect.any(Error),
+        });
+        return new Promise(() => {}); // Never resolves, simulating termination
+      });
```
ParidelPooya: Two problems with having the mock call `terminate` itself.

First, `safeDeserialize` is mocked out for this whole suite (`jest.mock("../../errors/serdes-errors/serdes-errors")`), so production code never reaches the real `terminate` call. The mock supplies the payload that the assertion at the bottom of the test then checks — the test is confirming its own mock, not the SDK.

Second, the payload doesn't match what the real helper produces. `serdes-errors.ts` calls `terminate({ reason, message })` with no `error` field, and the message is formatted as ``Deserialization failed for step "name" (stepId): ...`` rather than `"Failed to deserialize operation payload"`. So if `safeDeserialize` is ever un-mocked here, this expectation is wrong on both the shape and the string.

Also, `expect.any(Error)` on line 592 is being used as a plain value inside a mock implementation rather than in a matcher position. It only compares equal to the `expect.any(Error)` in the assertion because Jest's `equals` short-circuits when *both* sides are asymmetric matchers and falls back to structural comparison of the two `Any` objects — i.e. that field is tautological.

packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.test.ts
```diff
@@ -565,4 +566,119 @@ describe("WaitForCondition Handler", () => {
...
+      // up to the termination point, then assert termination was requested.
+      void handler(checkFunc, config);
+      await Promise.resolve();
+
+      expect(mockContext.terminationManager.terminate).toHaveBeenCalledWith({
+        reason: TerminationReason.SERDES_FAILED,
+        message: "Failed to deserialize operation payload",
+        error: expect.any(Error),
+      });
+      expect(checkFunc).not.toHaveBeenCalled();
```
ParidelPooya: Following on from the mock comment above: the contract this PR actually changes is *"the STARTED/READY restore path delegates to `safeDeserialize` with the right arguments"* — the terminate-on-failure behavior is already `safeDeserialize`'s own tested responsibility. Asserting the delegation directly is both honest and a stronger pin, and it lets the mock just return a never-resolving promise without pretending to terminate:

```ts
expect(mockSafeDeserialize).toHaveBeenCalledWith(
  expect.anything(),            // serdes
  JSON.stringify("checkpointed-state"),
  stepId,
  undefined,                    // name
  mockContext.terminationManager,
  mockContext.durableExecutionArn,
);
expect(checkFunc).not.toHaveBeenCalled();
```

I verified this variant still fails against the old `try/catch` implementation, so it keeps the regression coverage.

(If you take this, the `TerminationReason` import added at the top of the file becomes unused and will trip lint — it can come out too.)

packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.test.ts
```diff
@@ -565,4 +566,119 @@ describe("WaitForCondition Handler", () => {
...
+
+      // Force safeDeserialize to fail for state deserialization
+      mockSafeDeserialize.mockImplementationOnce(async () => {
+        mockContext.terminationManager.terminate({
+          reason: TerminationReason.SERDES_FAILED,
+          message: "Failed to deserialize operation payload",
+          error: expect.any(Error),
+        });
+        return new Promise(() => {}); // Never resolves, simulating termination
+      });
```
ParidelPooya: Same as the STARTED case above — the mock supplies the payload that lines 676-680 then assert, so this half of the test doesn't verify SDK behavior. Same suggested fix applies here.

ParidelPooya (commented):
The core fix is correct and I'd be happy to see it go in. Notes from verifying it locally, plus one thing to address on the tests and one missing changelog entry.

### What I verified

- `tsc --noEmit` clean; full SDK unit suite green (101 suites / 1210 tests).
- **Mutation check:** I reverted the handler change back to the old `try/catch` and re-ran only the two new tests — both fail. So they genuinely pin the changed branch, which matches @yaythomas's assessment.
- After this change, `serdes.deserialize` is no longer called directly anywhere in `src/` outside `safeDeserialize` itself. That's a consistency win beyond the bug fix: the `SUCCEEDED` replay path in this same file already used `safeDeserialize`, so both restore paths now behave identically.

The behavior change is the right call — silently resetting to `initialState` discards checkpointed progress and re-runs the condition loop from scratch, which is strictly worse than terminating the invocation. Termination is retryable at the service level, so a transient custom-serdes failure isn't permanently fatal.

packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.ts
yaythomas: Agreed — and thanks for running the mutation check independently. Nothing to change here.

packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.test.ts
yaythomas: Good catch on all three: the mock was authoring the payload it then asserted, the shape and message did not match what the real helper produces, and `expect.any(Error)` was tautological. Rewritten.

packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.test.ts
yaythomas: Taken as suggested. Verified both tests still fail against the old `try/catch`, so the regression pin holds. `TerminationReason` import removed.

One small note: it would not actually have tripped `npm run lint` — the root eslint config does not match SDK sources ("File ignored because no matching configuration"). Dead either way.

packages/aws-durable-execution-sdk-js/src/handlers/wait-for-condition-handler/wait-for-condition-handler.test.ts
yaythomas: Same fix applied here.

yaythomas:
All four addressed, plus the changelog entry — pushed as `252b50fa`, then merged current `main` in (`8204b8b9`), so the branch is level with main.

- Tests now assert the delegation to `safeDeserialize` rather than a payload the mock invented; both still fail against the old `try/catch`.
- `CHANGELOG.md` gains a `### Fixed` entry under `[Unreleased]` for the behaviour change.
- Node 22 and 24: build, lint, typecheck and the SDK suite (1210 tests) green. The built artifacts are byte-identical to the tree I ran the cloud suites against, so those results still stand: all ten conformance suites (171 tests, 0 failures) and full integration on 22.x and 24.x.

ParidelPooya (approved):

PR #774, opened 2026-07-29
fix(otel): durable.operation.status state machine (STARTED/terminal)

SilanHe:
> **Stacked PR (gh-stack style).** Review/merge bottom-up. This stack fixes the JS OTel invocation-view parity gaps surfaced by aws/aws-durable-execution-conformance-tests#40.
>
> 1. #773 — Workflow root span in all provider modes; invocation-rooted + Workflow links _(base: `main`)_
> 2. #774 — `durable.operation.status` state machine _(base: #773)_
> 3. #775 — wall-clock timing so op/attempt spans nest in Invocation span _(base: #774)_

---

## Summary

Adds the `durable.operation.status` state machine required by the updated invocation-view OTel contract (aws/aws-durable-execution-conformance-tests#40). Every operation carries `durable.operation.status` but not necessarily the attempts.

## Change

- Operation spans are stamped `STARTED` at span creation.
- On completion, terminal operations (STEP / WAIT / CHAINED_INVOKE / CALLBACK) overwrite it with their terminal status (`SUCCEEDED` / `FAILED`); **container CONTEXT operations** (child context, parallel, parallel branch) keep are marked SUCCEEDED.
- A suspended, never-resumed operation keeps `STARTED` (UNSET span status); the resumed terminal re-emission (continuation span) carries the terminal status.
- **Attempt spans** do not report `durable.operation.status = STARTED`; the attempt's success/failure is carried by `durable.attempt.outcome`.

## Conformance impact

Resolves `durable.operation.status: property is missing` (SUCCEEDED steps, attempt spans) and `expected 'STARTED'` (container child-context / parallel / map) across otel-invocation cases 1-9, 12-16.

## Testing

- `npm test` (otel package) — 183/183 pass (3 new)
- `tsc --noEmit` — clean

PR #775, opened 2026-07-29
fix(otel): wall-clock timing so op/attempt spans nest in Invocation span

SilanHe:
> **Stacked PR (gh-stack style).** Review/merge bottom-up. This stack fixes the JS OTel invocation-view parity gaps surfaced by aws/aws-durable-execution-conformance-tests#40.
>
> 1. #773 — Workflow root span in all provider modes; invocation-rooted + Workflow links _(base: `main`)_
> 2. #774 — `durable.operation.status` state machine _(base: #773)_
> 3. #775 — wall-clock timing so op/attempt spans nest in Invocation span _(base: #774)_

---

## Summary

Times operation/continuation/attempt spans with wall-clock so they nest within the (wall-clock) `Invocation` span, fixing the parent-timing envelope the conformance suite enforces.

## Change

Operation, continuation, and attempt spans were started/ended with the durable start/end timestamps. Because the `Invocation` span is a wall-clock span for the current invocation and these spans parent to it, a durable timestamp that predates this invocation (e.g. an operation first created in an earlier invocation of the same execution) pushed the child outside its parent's `[start, end]` window.

Time these spans with the tracer's current time instead, matching the Python/Java reference plugins. Only the root `Workflow` span keeps a backdated start (it is parentless, so no envelope applies).

## Conformance impact

Resolves `child span … starts before parent Invocation starts` / `ends after parent ends` (otel-invocation cases 2, 11, 15, 18).

## Testing

- `npm test` (otel package) — 184/184 pass (1 new envelope test)
- `tsc --noEmit` — clean


packages/aws-durable-execution-sdk-js-otel/src/invocation-plugin.ts
```diff
@@ -231,6 +231,15 @@ export class InvocationOtelPlugin implements DurableInstrumentationPlugin {
...
   async onOperationStart(info: OperationInfo): Promise<void> {
+    // Operation, continuation, and attempt spans are timed with wall-clock
+    // (the tracer's current time) rather than the durable start/end
+    // timestamps. The Invocation span is a wall-clock span for the current
+    // invocation, and these spans parent to it; using durable timestamps
+    // (which can predate this invocation's start, e.g. an operation created in
+    // an earlier invocation of the same execution) would push a child outside
+    // its parent's [start, end] window and break the parent-timing envelope
+    // the conformance suite enforces. Only the root Workflow span keeps a
+    // backdated start (it is parentless). Matches the Python/Java reference.
```
zhongkechen, line 252: FYI: Conformance tests also assert attempt spans in execution view to be contained in their invocation spans.

zhongkechen (dismissed):

packages/aws-durable-execution-sdk-js-otel/src/invocation-plugin.ts
SilanHe: ~how come?~ oh by contained you mean the start time and end time of the attempt spans are within the start and end times of the invocation spans.

ayushiahjolia (approved):

PR #781, opened 2026-08-02
feat(sdk): workflow model and durable-function codegen

ParidelPooya:
Two new packages: the model for a visual workflow, and the code generator that turns one into a durable Lambda handler.

 ``` 
 ┌───────────────────────┬───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Package               │ Files │ Role                                                                                                │
  ├───────────────────────┼───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ visual-workflow-model │ 28    │ Shared .dar model, JSON schema, identifier rules, migration — so authoring and codegen cannot drift │
  ├───────────────────────┼───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ -cdk                  │ 53    │ .dar → durable handler codegen, plus the DurableWorkflowFunction construct                          │
  └───────────────────────┴───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────┘
```
  This is the half that decides what actually runs in Lambda. The Workflow Studio UI and the desktop app build on it and follow in a stacked PR (feat/insight-workflow-studio),
  split so this half can be approved at line-by-line depth.

  90 files, +21,035. One file is a third of that and is generated, not hand-written: src/generated/apiDirectory.json (+6,648).

  Six existing files change

  84 files are new; 6 are modifications, all shared infrastructure:

  - .github/workflows/build.yml — the artifact glob was packages/**/dist*/*, which cannot match nested build output because a trailing * does not cross a /. This package's
  dist/generated/apiDirectory.json is the only such file in the repo, and unit-tests.yml downloads the artifact rather than rebuilding — so it was absent at test time,
  dist/apiVendors.js failed to load, and the entire cdk suite failed to run. Now dist*/**.
  - .github/workflows/iterate-publish-npm.sh — adds both packages, visual-workflow-model before -cdk (dependency order).
  - .github/workflows/lintcommit.js — adds the cdk scope.
  - .gitignore — restores **/.durable-executions-local/ to cover all emulator output rather than only its SDK cache; adds this package's cdk.out / cdk-outputs.json /
  deploy-verify output. Also ignores the built webview bundles and the SDK reflection cache — the monorepo shares one working tree, so those must be ignored on every branch or
  a git add -A sweeps in ~280k lines of Monaco worker output.
  - root package.json / package-lock.json — wires both packages into test and build.

  Read this first: DAG mode is present but gated off

  The most surprising thing in the diff. There is a complete second execution model here — DAG/fan-out authoring, codegen, and ~950 lines of tests — and it is disabled, because
  it generates calls to an SDK runtime that does not exist yet: context.dag(...), the dag.* task builders, .after(), .triggerRule(), and the DagResult. (NestingType happens to
  exist; nothing else on that list does.)

  Nothing downstream catches this. NodejsFunction bundles with esbuild, which transpiles without typechecking, so synth and bundle both succeed and the failure appears only
  when the deployed function is invoked, as TypeError: context.dag is not a function. The DAG tests assert on generated strings, so they stay green too.

  So codegen refuses by default. The gate lives in emitDagRegistrations — the sole function that emits registrations, reached by all three call sites, with options threaded
  through Scope on every recursion rather than a module global. Enable deliberately with allowDagMode.

  Four DAG arms additionally refuse rather than emit code that runs and quietly uses a task handle where a value belongs: a map iteratee and a parallel branch receive no deps
  parameter (their signatures are inherited from the SDK's existing map/parallel, so there is nowhere to bind one without inventing runtime API), a dynamic wait duration reads
  the handle, and onError cannot be honored where the closure does not wrap the failing operation.

  When the runtime lands, delete the gate and add the test that would have caught this in the first place: one that invokes a generated handler against the real SDK.

  Emission is validated by parsing

  Anything interpolated into generated code from a short inspector field must parse as exactly the construct it is dropped into — inputType, resultType, clientClass, command,
  errorType, runIf, shouldComplete, the four value fields (input, payload, initialState, startInput), the httpCall url/query/headers/body, and the SDK client package
  (restricted to @aws-sdk/client-*, since an unchecked specifier is a module esbuild will resolve and bundle).

  A step's or inline node's code stays raw by design — that is a whole function body the user is writing. The distinction that matters: a payload hidden in a one-line field is
  invisible to someone reviewing the canvas.

  One honest limit: requireExpression admits a parenthesized comma sequence, because that is one expression. This buys consistency across those fields, not a stronger guarantee
  than any one of them.

  Inferred IAM permissions fail closed

  Actions are inferred by pattern-matching step code, which cannot know which bucket or table a call targets. Any statement with a * in a resource — including a chainInvoke ARN
  like arn:aws:lambda:*:*:function:* — is withheld and reported as a CDK warning naming the exact actions, behind an explicit grantWildcardPermissions. The analyzer's
  under-grant warnings surface too, so a permission the analysis could not attribute is visible at synth rather than discovered as a runtime AccessDenied.

  Verification

  403 tests across 42 suites (visual-workflow-model 10/68, -cdk 32/335), wired into the root test and build scripts.

  Notably including a suite that compiles a generated handler and runs it against a mock DurableContext — asserting step order, results threaded between nodes, the event
  arriving as input, wait durations evaluated, condition branch selection, and that an error-route rejoin tail executes exactly once. That suite exists because every other test
  asserts on generated strings, which is how several defects stayed invisible; it found two bugs during review.

  aws-cdk-lib is pinned ^2.235.0 in all three places (peer, dev, README). durableConfig first shipped in 2.235.0 — on an earlier 2.x, CDK silently drops the unknown property,
  synth succeeds, and a non-durable Lambda deploys with no error until invoke.

  Known gaps

  Documented in code rather than glossed:

  - The construct embeds the legacy JSON .dar and never calls generateHandlerWithMap, so a construct-deployed function reopens at whole-node instead of statement granularity
  and ships no source map. Closing it means moving the .dar.ts serializer into visual-workflow-model — a cross-package refactor.
  - DAR_JSON_SCHEMA is advisory, not enforced. parseWorkflow checks only that nodes/edges are arrays; nothing validates against the schema, so its required lists and enums
  describe the format without rejecting a document. Marked as such rather than described as authoritative.
  - apiDirectory.json has no CI freshness or shape check, and is parsed eagerly at module load.
  - Several regexes (refuseUnbindableUpstreamRefs, COMMAND_RE, nodeReferencesSource) match identifiers inside comments and string literals. Fixing properly means AST-walking
  each; in nodeReferencesSource it silently changes emitted DAG code, so it deserves its own change with its own tests.
  - Neither package has an eslint config or lint script, no tsc --noEmit in test, no prepublishOnly, and their tsconfigs do not extend the root — divergences from siblings.
  - retentionPeriod is omitted from durableConfig when unset, so the documented 14-day default is not enforced by the construct; bundling.target is hardcoded node22 while
  runtime is overridable.

### CODE REVIEW

File: packages/aws-durable-execution-sdk-js-cdk/src/analyzePermissions.ts [hunk 1]
```diff
@@ -0,0 +1,332 @@
...
+  const collected: Collected = { snippets: [], invokes: [], jobs: [] };
+  collect(workflow, collected);
+
+  const actionsByService = new Map<string, Set<string>>();
+  const warnings = new Set<string>();
+
+  for (const code of collected.snippets) {
+    // Map command -> service via destructured import/require when possible.
+    const cmdService = new Map<string, string>();
+    for (const m of code.matchAll(IMPORT_RE)) {
```
yaythomas, line 195: doesnt strip comments.

A step containing // new DeleteTableCommand({}) plus a commented import still yields a dynamodb:DeleteTable grant attached to the role (concrete-resource path attaches without any opt-in; the wildcard gate only covers * resources).

maybe strip // and /* */ before scanning?

PR #788, opened 2026-08-04
refactor(sdk): own the durable execution wire types

ParidelPooya:
The SDK imported the durable execution protocol shapes -- Operation, OperationUpdate, ErrorObject, the checkpoint/state request and response types, and the OperationType/OperationStatus/OperationAction enums -- directly from @aws-sdk/client-lambda, across 28 source files. That made the SDK's own type surface a function of the AWS SDK's, and it meant the Lambda client was loaded at module scope even for callers that inject their own client.

Durable functions currently run only on Lambda, but support for further compute types is planned. This change removes the type-level coupling so that the transport can be swapped later, while leaving today's behaviour on Lambda unchanged.

Declare the wire model in the SDK
- src/types/wire/ now declares the 22 protocol shapes and 3 enums the SDK uses. The enums are `as const` string maps, matching how the service model represents them, so the resulting literal-union types and all runtime comparisons are identical to before.

- wire-model.aws-sdk-parity.test.ts guards the declaration against drift. It imports @aws-sdk/client-lambda for types only, is unreachable from src/index.ts, and is checked by both `build:types` and ts-jest, so a divergence fails the build. Shapes the SDK sends must match in both directions; shapes it only receives need only be assignable from the service model. Every shape also asserts an identical key set, because assignability alone does not catch added or removed fields.

- @aws-sdk/client-lambda remains a runtime dependency: Lambda's runtimes ship a pinned older minor of the AWS SDK that may not contain the durable execution commands, so it still has to be installed. What changes is that a single module imports it instead of 28.

Normalize timestamps on the way in
- Operations reach the SDK by two transports that disagree on how to represent a timestamp. The Lambda invocation event is plain JSON parsed by the runtime, so timestamps arrive as ISO-8601 strings; responses fetched through the Lambda API are deserialized by the AWS SDK into Date instances. Both were typed as Date and merged into a single history, so the declaration was wrong for the event path -- confirmed by the captured histories under packages/aws-durable-execution-sdk-js-examples, which contain values such as "2026-07-13T22:11:27.127Z".

- The consequences were live. The otel plugin passed these straight into OpenTelemetry's `startTime`, which does not accept strings, while the insight plugin had grown a defensive `toEpochMs(ts: unknown)` that coerced Date, number and string alike.

- Operations are now normalized as they enter, at the two boundaries where they arrive: execution-context initialization (invocation event plus paginated GetDurableExecutionState responses) and the checkpoint response. Only the raw shapes -- WireOperation, WireStepDetails, WireWaitDetails and the API response types -- carry the string-or-Date union. Operation and everything downstream of it, including the plugin surface, carry real Dates, so the public shape is unchanged from before this commit.

Load the Lambda client lazily

- DurableExecutionApiClient is now the only module that touches @aws-sdk/client-lambda at runtime, and it loads it with a dynamic import rather than at module scope. The load is started from the constructor without blocking it, so it overlaps with reading execution state and running the handler up to its first durable operation.

- Measured on Node 22, time from process start until the client can issue a request is unchanged (median 97ms lazy vs 102ms eager, with a tighter spread), while the SDK's own module load drops from ~69ms to ~9ms. After 50ms of other initialization work, 2.7ms of the load remains to be awaited. The added await on each request costs 0.114us.

- A dynamic import is used rather than a lazy require: the package has no `exports` field, so `main` -- the CommonJS build -- resolves either way and shares one module-cache entry, and a require would need a lint suppression and would diverge between the CJS and ESM outputs.

- A compute that supplies its own DurableExecutionClient never constructs this class, and so does not load the AWS SDK at all.


yaythomas:
despite the scary +25.7k/−14.2k, the real diff is small and disciplined. ~17k of the additions are regenerated api-extractor reports, and 92 of the 181 files are mechanical import swaps in the testing package. Draft review below — nothing posted to GitHub.

excellent work! the description makes falsifiable claims and every one of them holds up against the code. Things I particularly liked:

1. The parity guard (wire-model.aws-sdk-parity.test.ts) is genuinely airtight for today's setup: the type assertions distinguish send-shapes (mutual assignability) from receive-shapes (one direction), the SameKeys assertions catch added/removed fields that assignability alone would miss, and the runtime toEqual on the enum objects catches value drift the type layer can't see. I verified it's reachable from both build:types (tests live under src/, which the tsconfig includes) and ts-jest. The scope note about narrowing it when a second backend lands is exactly the right framing.
2. Normalization happens at exactly the two boundaries claimed, execution-context.ts (initial state + pagination) and checkpoint-manager.ts (checkpoint response), and I confirmed no other non-test code touches WireOperation. The otel plugin's startTime bug gets fixed without touching the otel package at all, which is the sign the boundary was drawn in the right place.
3. The lazy load handles the CJS default-interop wrinkle, clears the memo on failure so a transient error doesn't poison the process, and has an unmocked test covering the real resolution path.

yaythomas (approved):

PR #795, opened 2026-08-06
feat(insight): run Workflow Insight in VS Code and as a desktop app

ParidelPooya:
Workflow Insight was reachable only from inside VS Code because its behavior and its host were the same file: extension.ts held the webview panel, the settings writes and the save dialogs alongside query generation, the agentic loop, read-only enforcement and drill-down. This introduces a host seam and adds a second host behind it, so the same tooling runs outside VS Code without a second copy of the logic.

  ExplorerSession    everything the Explorer does, with no host in it. Moved out
                     of extension.ts unchanged; reaches the outside world only
                     through HostPort.
  HostPort           the whole contract a host implements: post a message, read
                     config, write settings, save a file, show info, read/write
                     saved queries.
  configCore         the settings shape, defaults, coercion and credential
                     resolution, with no vscode import. config.ts keeps the one
                     genuinely VS Code part (readConfig) and re-exports the rest.
  copilotBridge      the Copilot provider is VS Code's Language Model API, the
                     one capability that cannot be shared. It sits behind a port
                     so llm.ts stays host-free and the desktop app fails with an
                     explanation instead of "Cannot find module 'vscode'".
  settingsKeys       the recognized keys, shared because the desktop app has no
                     extension manifest to derive them from.
  hostBridge         the renderer picks acquireVsCodeApi() or the Electron
                     preload bridge. Inbound needs no abstraction: both hosts
                     deliver host messages as window "message" events.

extension.ts is now an adapter: 1721 lines to 251, all of it panel lifecycle and VS Code API. The desktop package is the same, in Electron. Neither contains query logic.

Behavior in VS Code is unchanged, and its 194 existing tests still pass. Note that they were not running in CI at all -- the root test script skipped insight-vscode -- so this adds both host packages to it, which is also what makes the new guard tests below meaningful.

Hosts report what they can do, so the renderer never offers a provider it cannot reach. The capability is derived from whether a Copilot bridge is installed rather than declared per host, so it cannot disagree with the runtime path. The Settings dropdown builds its options from it, the AI disclosure stops naming Copilot as a place data goes, and a stored llmProvider of "copilot" -- easy to carry over from the extension -- loads as the default instead of a value the dropdown does not offer.

Guards, because the invariants here are easy to break silently:

  hostAgnostic.test.ts     walks the real import graph from the modules the
                           desktop bundles and fails if any reaches vscode,
                           printing the chain. ./config and ./configCore differ
                           by four characters, and the bundler's error names the
                           imported file rather than the import at fault. It
                           found two live leaks while writing this
                           (destinationTest.ts) and a gap in itself: bare
                           import "./x" was unmatched.
  settingsKeys.test.ts     asserts the shared key list still matches the
                           extension manifest, including which keys are
                           boolean/number, so a setting added to one is not
                           silently unsettable in the other.
  hostCapabilities.test.ts pins that capabilities track the installed bridge
                           rather than being cached or hardcoded per host.
  assetPath.test.ts        the desktop asset server's traversal guard.

Two deliberate host differences, both documented in the desktop README: Copilot is VS Code only, and naming a saved query is interactive there (showInputBox) but automatic in the desktop app, since Electron has no native text prompt and HostPort.promptForText is optional.

Desktop security posture: contextIsolation on, nodeIntegration off, sandbox on, webviewTag off; assets served from a private insight:// scheme registered standard and secure with a per-load nonce CSP and default-src 'none'; navigation and popups denied; every IPC message checked against the window's own main frame before it can reach the session. Settings merge key-by-key against the shared allowlist rather than storing the renderer's object, keeping unknown and inherited names out of the store. Credentials are never persisted -- the standard SDK provider chain, honoring awsProfile, same as the extension.

Verified by launching the app: the bundle loads over insight://, React mounts, and its "ready" reaches the session through the preload bridge and the frame check. That run also surfaced two real defects, both fixed: the traversal guard normalized a "/"-prefixed path before joining, which collapses ".." against the filesystem root and silently discarded every traversal (all six rejection cases in assetPath.test.ts failed before the fix), and the CSP omitted data: from font-src, so Cloudscape's inlined icon font was refused.

Also adds the "insight" commit scope, and derives the scope list in lintcommit's self-test from the set itself -- that expectation had already gone stale before "ci" and "otel", so the self-test was failing on main.

Not included: signing/notarization, so no desktop release job is wired up -- this is build-and-run-locally for now.


packages/aws-durable-execution-sdk-js-insight-desktop/src/host.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (canceled || !filePath) return undefined;
+    await writeFile(filePath, request.data);
+    return filePath;
+  }
+
+  showInfo(message: string): void {
+    // The renderer already renders `status` text, so an informational message
+    // surfaces in the same place it does in VS Code's webview rather than
+    // interrupting with a modal dialog.
+    this.post({ type: "status", text: message });
```
yaythomas, line 61: In VS Code, showInfo is a transient toast; here it lands in the status line, which persists until the next status/results event overwrites it. So after "Results saved to /path", the desktop status bar keeps saying that indefinitely, where the extension's toast would have faded. It's a reasonable choice — but it's a third host difference in behavior, and the README's "Differences from the VS Code extension" section lists only two. Please add a line there (or clear the status after a delay).

packages/aws-durable-execution-sdk-js-insight-desktop/src/main.ts
```diff
@@ -0,0 +1,231 @@
...
+ * Path resolution and the traversal guard live in assetPath.ts so they can be
+ * tested without an Electron runtime; see assetPath.test.ts.
+ */
+async function serveAsset(request: Request): Promise<Response> {
+  const url = new URL(request.url);
+  // Only our single authority is served. `insight://evil/webview.js` is a
+  // distinct origin that the CSP and the preload's targeted postMessage do not
+  // account for, so refuse it rather than serving the app under a second origin.
+  if (url.host !== HOST) {
+    return new Response("Not found", { status: 404 });
```
yaythomas, line 118: decodeURIComponent(url.pathname) throws URIError on malformed percent-encoding (e.g. insight://app/%zz), and a pathname resolving to a directory that exists (e.g. /. → the media root) makes readFile reject with EISDIR. Either way the handler's promise rejects and Electron fails the request — not exploitable (only our own bundle can issue these requests under the CSP), but a try/catch around the body returning a 400 would keep the failure inside the contract this function otherwise defines so carefully. nit, ok to leave.

yaythomas (approved):

PR #801, opened 2026-08-07
fix(testing-sdk): stop duplicating ExecutionSucceeded

ParidelPooya:
The local test runner recorded two ExecutionSucceeded events, and two
  ExecutionStarted events, for an execution whose result was too large to
  return inline. The service records one of each.

  Split out of #797.

  The bug

  When a handler's return value exceeds Lambda's response size limit, the SDK
  checkpoints it as an EXECUTION-typed SUCCEED update under a synthetic
  execution-result-<timestamp> id, then returns { Status: SUCCEEDED, Result:
  "" } from the invocation.

  The runner treated those as two separate completions:

  - registerUpdate didn't recognise the synthetic id, so it created a second
   EXECUTION-typed operation and emitted ExecutionSucceeded under that id.
  - handleCompletedExecution then finalized the execution's own operation from
  the invocation response, emitting ExecutionSucceeded again under the
  execution id. Re-emitting that operation with an extra event also defeated
  IndexedOperations' history de-duplication, so the whole event list was
  appended a second time — which is why ExecutionStarted was duplicated too.

  What the service does

  From GetDurableExecutionHistory against real Lambda:

  ExecutionStarted      Id: 8a76ceda-dcf9-3eb4-b693-6b45c7aac4d2
  StepStarted           Id: c4ca4238a0b92382
  StepSucceeded         Id: c4ca4238a0b92382
  InvocationCompleted   Id: None
  ExecutionSucceeded    Id: 8a76ceda-dcf9-3eb4-b693-6b45c7aac4d2

  Exactly one terminal event, keyed by the execution id and carrying the
  payload. The result checkpoint does not surface as its own event, and the
  invocation response is ignored because the execution is already terminal.

  The result is genuinely stored rather than echoed back from the response —
  get-durable-execution returns Result of length 6291407 with Status:
  SUCCEEDED even though the SDK returned Result: "".

  The change

  checkpoint-server/storage/checkpoint-manager.ts:

  - registerUpdate redirects any EXECUTION-typed update onto the execution's
  own operation, ignoring the id the caller supplied. An EXECUTION-typed
  update always refers to the execution itself.
  - updateOperation returns before merging the invocation response into the
  stored operation, so a contradictory late status can't overwrite terminal
  state. Previously, following a checkpointed SUCCEED with
  updateOperation(..., FAILED) left the operation FAILED against a recorded
  ExecutionSucceeded.
  - It still publishes the original, unmerged operation. Ignore the response's
  contents, not its arrival: the invocation only runs to completion once an
  update is published. Suppressing it leaves the handler returning but the
  invocation unfinished — no InvocationCompleted event and an empty
  getInvocations(). The published update carries no new events, so the
  terminal event isn't duplicated.

  test-runner/local/test-execution-orchestrator.ts:

  - handleExecutionUpdate treats an update-less notification for an
  already-terminal execution as a no-op instead of throwing "Operation update
  is missing for execution update". That's precisely what the publish above
  is, and the throw was reached on every oversized run — invisible only
  because the execution promise had already resolved. The error is kept for
  the genuinely unexpected case: a missing update on a non-terminal execution.

  The checkpoint server already modelled this contract on the SDK-facing path,
  where registerUpdates rejects updates after completion with "Invalid
  checkpoint token". The internal finalization path bypassed it.

  Result

  Measured on the same oversized execution:

  ```
  ┌─────────────┬────────────┬──────────────┬──────────────┬────────────┐
  │             │ ExecutionS │ ExecutionSuc │ InvocationCo │ getInvocat │
  │             │ tarted     │ ceeded       │ mpleted      │ ions()     │
  ├─────────────┼────────────┼──────────────┼──────────────┼────────────┤
  │ real Lambda │ 1          │ 1            │ present      │ 1          │
  ├─────────────┼────────────┼──────────────┼──────────────┼────────────┤
  │ before      │ 2          │ 2            │ present      │ 1          │
  ├─────────────┼────────────┼──────────────┼──────────────┼────────────┤
  │ after       │ 1          │ 1            │ present      │ 1          │
  └─────────────┴────────────┴──────────────┴──────────────┴────────────┘
  ```

  Only executions with more than one EXECUTION-typed operation are affected,
  which today means only the oversized-result path.

  Why it matters beyond the event count

  LocalOperationStorage builds history from the local runner, whereas
  CloudDurableTestRunner reads the service's authoritative event log via
  HistoryPoller, and assertEventSignatures grants no tolerance on
  ExecutionStarted or terminal events. Any test asserting one committed
  snapshot in both modes could only ever pass locally, and would fail against
  real Lambda for reasons pointing at the snapshot rather than at the harness.

  Known remaining difference

  The service emits InvocationCompleted before the terminal event; the local
  runner emits it after. Closing that needs EventId assignment shared between
  the checkpoint server and the invocation tracker, which can shift history
  order for every execution and may require regenerating example snapshots —
  so it's left to the event-ordering TODO already in
  local-durable-test-runner.ts and tracked separately. The assertions here are
  on counts and ids rather than positions.

  Testing

  Unit coverage in checkpoint-manager-execution-result.test.ts:

  - an EXECUTION-typed checkpoint under a foreign id lands on the execution
  operation, creating no second operation
  - the ignored response is published without being merged and without adding
  a duplicate event
  - a late contradictory status doesn't overwrite terminal state
  - the ordinary inline path still records its terminal event

  Integration coverage in oversized-execution-result.integration.test.ts,
  asserting the assembled history end to end — the layer whose absence hid
  this, since the duplicate events only appear once history is assembled. It
  pins all four columns of the table above, and fails before this change
  (duplicate ExecutionStarted) and if either half of the fix is reverted.

  Suites: testing package 71 suites / 1174 tests. Examples package unaffected:
  133 suites / 187 tests. tsc, ESLint and Prettier clean.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-testing/src/checkpoint-server/storage/checkpoint-manager.ts
+++ b/packages/aws-durable-execution-sdk-js-testing/src/checkpoint-server/storage/checkpoint-manager.ts
@@ -38,6 +38,19 @@ export class CheckpointManager {
 
   private _isExecutionCompleted = false;
 
+  /**
+   * Id of the execution's own EXECUTION-typed operation, assigned in
+   * {@link initialize}.
+   *
+   * The service keys the execution's terminal event by this id. When the SDK
+   * checkpoints an oversized result it invents its own id
+   * (`execution-result-<timestamp>`), but the service treats any EXECUTION-typed
+   * update as an update to the execution itself and ignores that id — a real
+   * execution history contains exactly one ExecutionSucceeded, carrying this id.
+   * {@link registerUpdate} redirects such updates here to match.
+   */
+  private executionOperationId: string | undefined = undefined;
+
   constructor(executionId: ExecutionId) {
     this.callbackManager = new CallbackManager(executionId, this);
   }
@@ -171,6 +184,7 @@ export class CheckpointManager {
     } satisfies OperationEvents;
 
     this.operationDataMap.set(initialId, initialOperationEvents);
+    this.executionOperationId = initialId;
 
     return initialOperationEvents;
   }
@@ -344,6 +358,12 @@ export class CheckpointManager {
             copied.operation.ContextDetails?.ReplayChildren,
         };
         break;
+      case OperationType.EXECUTION:
+        // The execution reached a terminal state through a checkpoint (the
+        // oversized-result path). Further checkpoints must be rejected, and the
+        // invocation response must not add a second terminal event.
+        this._isExecutionCompleted = true;
+        break;
     }
 
     return copied;
@@ -400,6 +420,26 @@ export class CheckpointManager {
 
     if (!newOperation.Status) {
       throw new Error("Missing Status in operation");
+    }
+
+    // The execution may already have been completed by an EXECUTION-typed
+    // checkpoint (the oversized-result path). The service ignores the invocation
+    // response in that case — the execution is terminal and its result is
+    // already stored — so return before the response is merged.
+    //
+    // Ignore the response's *contents*, not its arrival. Merging would let a
+    // contradictory status overwrite the terminal state, leaving a stored FAILED
+    // against a recorded ExecutionSucceeded. But the update must still be
+    // published: the invocation only runs to completion once an update is
+    // published, and without it no InvocationCompleted event is recorded and
+    // `getInvocations()` comes back empty. Publishing the original operation is
+    // a no-op for history, since it carries no new events.
+    if (
+      operationData.operation.Type === OperationType.EXECUTION &&
+      this._isExecutionCompleted
+    ) {
+      this.addOperationUpdate({ ...operationData, update: undefined });
+      return operationData;
     }
 
     const newOperationData: OperationEvents = {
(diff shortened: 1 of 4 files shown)


packages/aws-durable-execution-sdk-js-testing/src/checkpoint-server/storage/checkpoint-manager.ts
```diff
@@ -476,6 +496,14 @@ export class CheckpointManager {
         break;
       }
       case OperationType.EXECUTION: {
+        // The execution may already have been completed by an EXECUTION-typed
+        // checkpoint (the oversized-result path). The service ignores the
+        // invocation response in that case — the execution is terminal and its
+        // result is already stored — so do not record a second terminal event.
+        if (this._isExecutionCompleted) {
```
yaythomas: break only exits the switch but the invocation response has already been merged into newOperationData, and execution continues to addOperationUpdate at line 537.

Therefore, the supposedly ignored response is still stored and published.

Consequences:

1. A real oversized LocalDurableTestRunner execution published a second EXECUTION update with update: undefined, causing handleExecutionUpdate to throw Operation update is missing for execution update. This happened on all 10 runs; the error is hidden because the execution promise has already resolved.
2. Following a checkpointed SUCCEED with updateOperation(...FAILED) produced stored status FAILED while history remained ExecutionStarted, ExecutionSucceeded.

Suggest return the existing operationData before constructing newOperationData or calling addOperationUpdate. Add coverage asserting no second pending update is emitted and a late contradictory status cannot overwrite terminal state.

yaythomas (changes requested):

packages/aws-durable-execution-sdk-js-testing/src/checkpoint-server/storage/checkpoint-manager.ts
ParidelPooya: Addressed

yaythomas (approved):
The oversized-result path now: (a) redirects synthetic checkpoint IDs to the execution's own ID, (b) prevents merge of late invocation responses, (c) still publishes to complete the invocation, and (d) gracefully handles the resulting update-less notification. Unit tests pin each invariant; the integration test validates the end-to-end flow.

PR #809, opened 2026-08-08
refactor(insight): extract durable-insight-core from the VS Code package

ParidelPooya:
The Electron host reaches its shared code by relative path across a package boundary,
  with no declared dependency:

      from "../../aws-durable-execution-sdk-js-insight-vscode/src/explorerSession"

  Tolerable with two hosts. An MCP host is coming, and at three this becomes a package
  importing across two siblings to reach code that lives, misleadingly, inside the *VS
  Code* package. Cheaper to fix now than after a third consumer exists.

  No behavior change to either host.

  ## What moved, and how it was decided

  Measured rather than assumed: only `extension.ts` and `config.ts` import `vscode`, so
  all 25 remaining modules move to the new `@aws/durable-insight-core`.

  The desktop host transitively needs all 25 — which is why the LLM modules come along
  even though the MCP host won't use them: `ExplorerSession` reaches `llm` and
  `agentLoop`. No module was reachable from neither host, so nothing was left stranded.

  ## Enforcement, and four holes review found in it

  A one-time cleanup plus an assertion in a PR description doesn't stop the next sideways
  import. Getting the enforcement right took four rounds, each found by injecting a fault
  or running a scanner rather than by reading code.

  **1. The rule was named too narrowly.** The first guard checked only `vscode`. An
  `electron` import in a core module that no test happened to cover passed *everything* —
  eslint, `typecheck:hosts`, both bundles, all 251 tests. Nine of core's 25 modules have
  no direct test, so there was no signal at all. The real invariant is that shared code
  must reach **no host's API**:

  | Module | Only exists in |
  | --- | --- |
  | `vscode` | the VS Code extension host |
  | `electron` | the desktop app |
  | `@modelcontextprotocol/sdk` | the MCP server |

  One detector (`hostModuleScan.ts`) with **65 self-tests** covers every module in every
  import form — static, bare side-effect, `require`, dynamic `import`, and **subpaths**.
  Subpath matching is load-bearing, not defensive: the MCP SDK is only ever imported that
  way, so an exact-specifier match would have missed every real usage while looking
  correct. A test asserts the module list still has all three, so a fourth host can't be
  added silently.

  **2. Per-file scans can't see transitively.** A file that reaches sideways into a sibling
  contains no host import of its own while pulling one in through the back door —
  confirmed: a core file importing `"../../...-insight-vscode/src/config"` passed all 277
  tests, typecheck, and both bundles. `findEscapingImports` resolves each relative
  specifier against the importing file's directory and rejects anything landing outside the
  package root. Depth-aware rather than a `../../` string match, since the same specifier
  escapes from one depth and not another — which is its own test case.

  **3. The original defect had no CI gate.** eslint ran only via lint-staged on
  pre-commit, which `--no-verify` bypasses. **`npm run lint` is now a step in the
  `insight-hosts` job**, so the boundary rules gate a pull request. That required
  extending the ignore list, because `eslint .` was linting generated `dist-cjs` and
  `.rollup.cache` JavaScript.

  **4. The new detector shipped a ReDoS.** CodeQL flagged `js/polynomial-redos` (high) on
  `replace(/\/+$/, "")` — a greedy quantifier anchored at the end. Measured, it is worse
  than "may run slow": 170ms at 10k slashes, 2.7s at 40k, 43s at 160k, **174s at 320k**,
  exactly quadratic. See below.

  Every package now runs the mirror guard — "my own host API and no other" — so the
  extension host is stopped from importing `electron`, the desktop app from importing
  `vscode`, and core from importing any. The VS Code package additionally guards its
  **renderer** separately: `webview-ui` runs in a browser and may import none of the three,
  since it reaches the host through the `acquireVsCodeApi()` global rather than by
  importing the API. The detector previously existed as two copies, one with no self-tests
  at all.

  ## Three ReDoS regexes removed

  Grepping for the flagged shape found two more, both in files this PR **moved** rather
  than wrote, both on the same 16×-per-4× curve:

  - **`llm.ts`** normalized the local-server URL with the identical regex. Its input is a
    user-configured setting (`localServerUrl`), making it the most exposed of the three.
  - **`queryShape.ts`** trimmed trailing whitespace with `replace(/\s+$/, "")` on a
    generated query string.

  All three are fixed. Leaving two known-quadratic patterns in code this PR just relocated,
  after a scanner flagged the third, isn't a defensible place to stop. The whitespace one
  becomes `trimEnd()` — native, linear, identical in effect; the two slash strips share one
  helper. Equivalence checked across eleven inputs including empty, all-slashes, and
  interior-slash cases.

  Two further candidates were **measured and left alone** because they're linear:
  `/^\s*\*\s*$/` in `queryShape.ts` (anchored both ends with a literal between the
  quantifiers) and the anchored character-class patterns in the MCP package. A blanket
  sweep would have churned them for nothing.

  The regression guard is a pathological-input test with a deliberately generous 1s bound —
  it asserts "not catastrophic" rather than benchmarking, so it won't flake on a slow
  runner. Reintroducing the regex makes it take 68.7 seconds and fail. CodeQL remains the
  standing gate for the class; it's what found this one.

  ## The release workflow could no longer install the extension

  `vscode-extension-release.yml` can't run `vsce` from inside the workspace, so it copies
  just the VS Code package to a temp directory and runs a standalone
  `npm install --no-workspaces`. That resolves from the public registry — and this package
  now depends on `@aws/durable-insight-core`, which is `"private": true` and never
  published. Reproducing the isolate step gives `E404`. **The workflow is
  `workflow_dispatch`-only, so no pull request runs it and the checks stayed green.**

  The workflow now packs private workspace deps before that install. The tarball-vs-
  directory distinction matters: `file:<dir>` is link semantics — npm creates the symlink
  but installs **none** of the linked package's dependencies, so core's 16 AWS SDK deps
  went missing and esbuild failed to resolve them. Verified end to end: pack → rewrite →
  `npm install --no-workspaces` with zero E404s → webview build → esbuild producing a
  2.24 MB bundle → `getLlama()` resolving.

  `src/packaging.test.ts` is the standing guard, since the workflow itself can't run on a
  PR. It asserts every private workspace dependency is staged, and that staging uses a
  tarball. Removing the staging loop fails it; adding a second private dependency without
  staging fails it by name.

  ## Decisions worth knowing

  **Core is `private: true`, scoped, `main` at `src/index.ts`, no build step.** Both
  consumers bundle from TypeScript source with esbuild; publishing would turn an internal
  boundary into a semver commitment for no external benefit. Wired into root `test` and
  `typecheck:hosts` but deliberately **not** `build`, with `noEmit` set so a stray `tsc`
  can't leave an untracked `dist`.

  **Scoped even though private,** because the unscoped name was unclaimed on the public
  registry and consumers referenced it as `"*"`. Workspace resolution always wins in-repo,
  so this is hardening — but nothing else can publish under `@aws/`, and consumers now pin
  the exact version.

  **`src/index.ts` is a barrel,** safe because no two modules export the same name.

  ## Verification

  - Tests **222 + 36 → 327 + 33 + 48**. The seven old `hostAgnostic` cases are the only
    removals.
  - `npm run lint`, `typecheck:hosts`, renderer typecheck, both bundles, root `npm test`,
    `lintcommit` 30/30 — all from a fresh `npm ci` in an isolated worktree.
  - Every guard mutation-verified: each host module, in each package, in each import form,
    fails the specific named case and only that case.
  - The isolated VSIX packaging path verified end to end, since no PR exercises it.
  - The desktop app **launches and runs a query** against the extracted core.
  - `tsc` in core produces no `dist`.

  ## Notes for the reviewer

  - **39 of 66 files are R100 renames** (byte-identical) and render as moves. Two more are
    R099/R098: `llm.ts` and `queryShape.ts`, differing only by the ReDoS fixes above. Those
    are the only moved files with any content change, and both are one line plus comments.
  - **Two test files deliberately stayed** in `-insight-vscode` because they read files
    that stay there: `settingsKeys.test.ts` asserts `SETTING_KEYS` matches the extension
    manifest, and `hostCapabilitiesContract.test.ts` reads `webview-ui`.
  - **Pre-existing, unchanged:** npm workspaces hoist `node-llama-cpp` to the root while
    `.vscodeignore` un-ignores a package-relative path. Identical on `main`, and
    `require.resolve` still succeeds, so VSIX packaging is unaffected. The desktop package
    now at least *declares* the dependency its esbuild config marks external.
  - **Version skew is intentional:** core is `0.1.0-alpha.0` because it's new.
  - `CONTRIBUTING.md` lists five valid scopes; `lintcommit.js` enforces nine including
    `insight`. Pre-existing drift, not addressed here.
  - Every real gap here was found by **injecting a fault or running a scanner**, not by
    reading code — the bare side-effect import, the `electron` blind spot, the MCP SDK
    subpath, the transitive escape, the packaging E404, and the ReDoS. When reviewing the
    guards, the useful question about each is "what makes it fail?", not "does it pass?".

  ## Testing (per CONTRIBUTING)

  Unit tests throughout, redistributed with the code they cover, plus 65 self-tests for the
  boundary detector, a packaging invariant test, and a ReDoS regression test. No
  conformance tests added. Not an SDK feature or API change, so no examples needed.


packages/aws-durable-execution-sdk-js-insight-vscode/package.json
```diff
@@ -269,22 +269,8 @@
...
-    "@aws-sdk/client-rds-data": "^3.658.0",
-    "@aws-sdk/client-redshift-data": "^3.658.0",
-    "@aws-sdk/client-sqs": "^3.658.0",
-    "@aws-sdk/credential-providers": "^3.658.0",
-    "@aws-sdk/util-dynamodb": "^3.658.0",
-    "@jitl/quickjs-singlefile-cjs-release-sync": "^0.32.0",
-    "@smithy/signature-v4": "^5.0.0",
-    "node-llama-cpp": "^3.18.1",
-    "quickjs-emscripten-core": "^0.32.0"
+    "@aws/durable-insight-core": "0.1.0-alpha.0",
```
yaythomas, line 272: The isolated VSIX packaging path can no longer install this package. It copies only the VS Code package, but that package now depends on the private, unpublished @aws/durable-insight-core.

Consequently, npm install --no-workspaces attempts to resolve core from npm and returns E404. This workflow is dispatch-only, so the green PR checks do not exercise the failure.

If this stages core into the isolated build or bundle core before isolation and adjust the packaging manifest, then add a packaging smoke test, that tests it.

packages/aws-durable-execution-sdk-js-insight-desktop/src/hostAgnostic.test.ts
```diff
@@ -0,0 +1,76 @@
...
+describe("desktop package imports only its own host API", () => {
+  // Guards against the guard passing by finding nothing.
+  it("finds its own sources, including known members", () => {
+    expect(names.length).toBeGreaterThan(3);
+    for (const expected of ["main.ts", "host.ts", "settings.ts"]) {
+      expect(names).toContain(expected);
+    }
+  });
+
+  it.each(names)("%s imports no other host's API", (name) => {
```
yaythomas, line 76: the original cross-package relative-import failure still has no CI enforcement. no-restricted-imports catches it locally, but eslint does not run in CI. The new hostAgnostic tests only call findHostModules on each file, so a sibling-relative import and any host API reached transitively through it passes.

suggestion: either run the boundary eslint rules in CI or extend the tests to reject imports that escape into sibling packages.

packages/aws-durable-execution-sdk-js-insight-vscode/package.json
ParidelPooya: Addressed

packages/aws-durable-execution-sdk-js-insight-desktop/src/hostAgnostic.test.ts
ParidelPooya: Addressed

yaythomas (approved):
VSIX packaging: Fixed with a packed core tarball. I independently reproduced staging and installation of @aws/durable-insight-core@0.1.0-alpha.0, then built the isolated webview and 2.2 MB extension bundle successfully.
Boundary enforcement: Fixed in both CI lint and unit tests. Mutation probes confirmed that each now rejects the original sibling-relative import.
ReDoS: The three replacements are linear and behaviorally equivalent. CodeQL now passes.
Validation: Lint, all host typechecks, and tests passed locally: core 327, VS Code 33, desktop 48. All GitHub checks are green.

PR #810, opened 2026-08-09
feat(insight): serve Workflow Insight from an MCP server host

ParidelPooya:
Customers have asked to use Workflow Insight from the AI agent they already work in —
  Claude Code, Kiro, Cursor — rather than through a VS Code extension or a desktop app.
  The competitive analysis in this repo already notes DBOS shipping an MCP server, so this
  is parity rather than differentiation.

  The real win isn't packaging, it's the follow-up question. Today a user asks one thing,
  gets a table, and starts over. In an agent, *"show me the input payloads for those three"*
  is just the next turn — no new UI, no new query mode, no chart config. That's a capability
  the extension structurally cannot offer.

  Third host on the seam #795 introduced, built on the core extracted in #809. The design
  document ships in this change; § references below point at it.

  ## What it is

  A stdio MCP server, published so `npx -y @aws/durable-insight-mcp` works, exposing five
  tools over all seven queryable destinations:

  | Tool | Purpose |
  | --- | --- |
  | `test_destination` | Connectivity check; names unset env vars with no AWS call |
  | `describe_schema` | Per-destination record schema and query idioms, from core |
  | `list_executions` | Filtered listing with **no SQL required** — cheaper and unfailable |
  | `get_execution` | One execution in full, including `input`/`output` payloads |
  | `query` | The escape hatch for anything the structured tools can't express |

  Configuration is environment-only (`DURABLE_INSIGHT_*`, with `AWS_REGION`/`AWS_PROFILE`
  honored), and **credentials need no configuration at all** — the standard chain already
  works. A typical customer sets two to four variables, making this the easiest of the three
  hosts to configure.

  **It drops a subsystem rather than adding one.** The agent is the model, so ~1,700 lines of
  provider plumbing don't come along: no provider selection, no model lists, no on-device
  binary, none of the Electron packaging friction. Eight settings are deliberately rejected,
  with a contract test asserting those eight plus the 25 accepted exactly partition
  `SETTING_KEYS`, so neither list can drift from the extension's manifest.

  ## The security shape, which is where review should concentrate

  In the other hosts `assertReadOnly` guards SQL our own prompt produced. Here it guards SQL
  from a model we don't control, in a loop, for a user who may approve without reading.

  There is also a structural hazard, verified: **`assertReadOnly` and `ensureLimit` are called
  only inside `explorerSession.ts`, never inside the engine runners.** This host deliberately
  bypasses `ExplorerSession` (§6.3), so the runners are reachable with no read-only
  enforcement — nothing in core would refuse a `DELETE`.

  The mitigation is deliberately not a call at each site, which is the pattern that let it be
  forgotten. **`readOnlyQuery.ts` is a single choke point** and `queryChokePoint.test.ts`
  mechanically asserts nothing else imports core's query surface.

  Three things worth knowing about the tests:

  - **Rejection is proven by asserting the runner mock was never invoked**, not merely that an
    error was thrown. A rejection that still reached AWS would be worthless. Covered for every
    engine: all 14 write keywords, multi-statement, comment-obfuscated, and both
    data-modifying CTE forms.
  - **Acceptance is tested too**, because a suite that only checks rejection passes on a
    function that rejects everything: plain `SELECT`, read-only `WITH`, a keyword inside a
    string literal, `REPLACE()` as a scalar function.
  - **`assertReadOnly` is the backstop, not the sanitizer.** An injected `' OR '1'='1` is a
    valid SELECT and it will allow it. Athena cannot parameterize, so agent-supplied values
    are escaped in `sqlSafe.ts`, closed vocabularies are *rejected* rather than escaped, and
    the injection tests assert on the **generated SQL text** so a reviewer can see the payload
    stayed one literal.

  ### The exception that looks like a bug

  **CloudWatch Logs Insights is deliberately not guarded by `assertReadOnly`** (§6.7). It's a
  pipe language — `fields @timestamp | filter … | stats …` — so a `SELECT`/`WITH` prefix check
  would reject *every valid query*. Read-only comes from the API surface instead: the language
  has no write forms and `StartQuery` can only read. `explorerSession.ts` draws the same line.

  Because "add the guard here too, for consistency" is such a plausible future change, **the
  test is inverted**: it asserts valid pipe queries are *accepted*. Adding `assertReadOnly`
  there fails 22 tests.

  Two consequences: `ensureLimit` is correct for that engine and wrong for every other (it
  emits `| limit N`, a syntax error on Trino/PartiQL), and the **escaping differs** — Logs
  Insights literals are double-quoted, so core's `escapeQuotedString` is required.

  `sqs` is refused outright rather than half-supported: it isn't queryable
  (`explorerSession.ts:828`, and `sqs.ts` exposes only `listenToQueue`).

  ## Skill delivery, in three layers

  Tools give an agent reach; these give it competence. Without schema guidance an agent writes
  plausible-but-wrong queries across seven subtly different backends and concludes the tool is
  broken.

  1. **`describe_schema` as a tool** — works in every client, zero setup. The floor.
  2. **Two MCP prompts** — ship *inside* the server, so Kiro lists them under `/prompts`.
  3. **`SKILL.md`** (4 KB) — for clients with skill support, loaded on demand.

  **None contains a single destination-specific schema fact.** They delegate to
  `describe_schema`, which makes drift impossible rather than merely detectable — and avoids
  loading Athena's 10,143 characters of guidance for a DynamoDB user. `skillDrift.test.ts`
  asserts schema-owned tokens appear nowhere in the skill or prompts, **and first asserts each
  token genuinely occurs** in some `buildSystemPrompt` output, so the guard cannot pass
  vacuously. That check earned its place immediately: a seventh candidate token appeared in no
  prompt output and was dropped.

  ## Verified against real data, not just green tests

  Driven over real stdio against a live Aurora Serverless v2 cluster with 5,808 records:

  - `test_destination` passes both checks in **0.7s**
  - `list_executions` returns real rows with correct `snake_case` columns
  - `get_execution` returns a failed execution's error, operations array, and input payload
  - A `GROUP BY` aggregate runs: `SUCCEEDED 5594 / 2308ms avg`, `FAILED 214 / 5986ms avg`
  - **Six write attempts refused with Admin credentials in hand** — `DELETE`, a data-modifying
    CTE, a multi-statement `DROP`, `UPDATE`, `TRUNCATE`, a comment-obfuscated `DELETE` — with
    the row count unchanged afterwards

  That exercise found a bug no unit test could: `describe_schema` told the agent to call
  `emit_query`, a tool that exists in the *extension's* LLM loop and not here. The existing
  tests asserted the guidance was long and self-consistent, and it was — nothing in them knew
  its closing sentence addressed a different host.

  ## Review round: six silent failures

  Every one of these passed CI, typecheck, and the full suite. They're grouped because they
  share a shape — a guard that could not fail, or a promise with no implementation.

  **The package was never published.** `iterate-publish-npm.sh` publishes from a hard-coded
  list this package wasn't in, so `npx -y @aws/durable-insight-mcp` — the entry point the
  entire README is written around — could not have worked. `packaging.test.ts` read as coverage
  for exactly this and wasn't: it asserted `private === false`, a manifest flag that *permits*
  publishing rather than causing it. The new invariant is keyed on the **`@aws/` scope**, not
  on `private !== true`, because three packages here are non-private and correctly never
  published — the broader rule would fail on state that is already right.

  **`query` declared a `limit` it never read.** Validated against the cap, described to the
  model as the row maximum, and the handler destructured `{ sql, lookbackHours }`. An agent
  asking for ten rows got a thousand — the token cost the structured tools exist to avoid. Now
  threaded through as a per-call cap that can lower the bound but never raise it. Fixing it
  exposed **the same bug in a second place the review didn't reach**: `list_executions` on
  DynamoDB, whose PartiQL supports neither `ORDER BY` nor `LIMIT`, so its statement carried no
  bound at all. Since schema and handler are only comparable as source,
  `toolWiring.test.ts` parses the TypeScript AST and asserts in both directions that every
  declared parameter is read.

  **A misspelled `DESTINATION_TYPE` was silent.** `normalizeConfig` ends in a fallback to
  `cloudwatch-logs-exporter` — right for the extension, where the value comes from a dropdown,
  and a trap for an environment-only host where `dynamo` or `DynamoDB` is the expected mistake.
  It produced a server reporting `DURABLE_INSIGHT_LOG_GROUP_NAME` as missing to someone who had
  configured DynamoDB correctly, and `missingRequiredEnvVars` could never name the real problem
  because by then the typo was gone. Core now exports the destination list at runtime and
  `normalizeConfig` resolves through it, so the warning cannot drift from the parser.

  **DynamoDB results could be incomplete while reporting `truncated: false`.** One
  non-paginating `ExecuteStatement`, a ~1 MB service response limit, and the `NextToken` saying
  more existed was discarded — so 400 rows of a matching 5,000 came back as a complete answer,
  which an agent will state as one. Core surfaces `hasMore`, **required rather than optional**
  so no construction site can omit it and default to falsy.

  **The choke-point guard was named too narrowly.** It forbade the six `run*Query` runners and
  missed six more core exports that execute a query, most by calling a runner internally —
  including `ensureAthenaTable`, which runs DDL. A file importing `fetchAthenaRecord` executed
  SQL while mentioning no runner, so the guard stayed green over the hole it exists to close.
  This is the `electron` blind spot from #809 in a new costume, so the fix is structural rather
  than three more names: the set is **derived from core's AST**, a meta-test fails when a new
  query-executing export appears, and the two deliberate exceptions carry written reasons plus
  an assertion that each is still used.

  **`escapeSqlString` doubled quotes only.** Redshift's own `QUOTE_LITERAL` doubles quotes *and
  backslashes*, because backslash escapes in its literals — so a trailing backslash yields
  `'foo\'` with the closing quote escaped and the literal running on. Doubling backslashes would
  be wrong for Athena, where they're ordinary characters. Backslashes are **rejected** instead,
  which is lossless: only execution ARNs and Lambda function names reach that code.

  Plus five smaller ones: `get_execution` on a log destination inherited a 7-day window with no
  way to widen it and reported a miss identically to an absence (now takes `lookbackHours` and
  returns `searchedLookbackHours`); partition hints passed to a non-Athena destination are
  reported in `ignoredParams` rather than dropped; the `AWS_REGION` fallback reads
  `AWS_DEFAULT_REGION` too, matching the `normalizeConfig` it shadows; a second unread entry in
  `FOREIGN_TOOL_MARKERS` is gone; and the three per-engine `truncated` derivations are
  documented where a reader compares them.

  ## Notes for the reviewer

  - **26 of 37 files are a new, self-contained package.** Of the rest, three are core
    (**55 insertions across 3 files** — `hasMore`, the runtime destination list, and
    `normalizeConfig` resolving through it), two are `.github`, plus the lockfile, root
    `package.json`, and the design doc. Nothing here can regress the extension or the desktop
    app, and both suites are unchanged at 33 and 48.
  - Of ~7,000 lines of real code, **3,281 are tests** and 1,258 are documentation; the
    lockfile's 2,570 lines can be skipped.
  - **The `hasMore` pinning test lives in core, not here** — and that's deliberate. This
    package mocks the runner, so a version that stopped reading `NextToken` still satisfied its
    tests; I confirmed that by mutation before moving the test. Same reasoning as the
    stdout-purity test: the obvious place to assert it can't see the failure.
  - **`stdout` purity is asserted directly.** stdout *is* the MCP transport, so a stray
    `console.log` corrupts it — yet injecting one did **not** fail an end-to-end session test,
    because the SDK client catches per-line parse errors, reports them to an unset handler, and
    continues. Only asserting the spawned binary emits nothing on stdout catches it.
  - **The published package declares only `@modelcontextprotocol/sdk` and `zod`.** Core is a
    `devDependency` because esbuild inlines it — it was briefly a runtime dependency, which made
    `npx` fail with E404 for every customer.
  - Bundle is **1.85 MB**. `quickjs` is fully tree-shaken despite core's barrel re-exporting the
    sandbox; `node-llama-cpp` stays external because it's only ever dynamically imported.
  - **`runSandboxedJs` is deliberately not exposed.** Arbitrary agent-authored JavaScript is a
    different risk class from a bounded read-only query, and the agent has its own code
    execution.
  - **Not shipped, deliberately:** charts and the interactive grid have no home in a text
    protocol, and SQS tailing needs a design for a request/response transport (§11, Phase 6).
  - When reviewing the guards, the useful question about each is **"what makes it fail?"**, not
    "does it pass?" — every gap found in this work came from injecting a fault or running a
    scanner, never from reading code.

  ## Testing (per CONTRIBUTING)

  Unit tests throughout — **391** in this package, **334** in core (up from 326 and 329) — plus
  an integration test that spawns the built binary and speaks real MCP over stdio, contract
  tests for the settings partition and the skill anti-drift rule, a packaging invariant test,
  and the AST-derived choke-point meta-guard. Every guard is mutation-verified: break the thing,
  watch the specific named case fail, restore. No conformance tests added. Not an SDK feature or
  API change, so no examples needed.

### DIFF
--- a/packages/durable-insight-mcp/src/tools.ts
+++ b/packages/durable-insight-mcp/src/tools.ts
@@ -36,6 +36,7 @@ import {
   type InsightConfig,
 } from "@aws/durable-insight-core";
 import {
+  DEFAULT_LOG_TIME_RANGE_MS,
   LOGS_INSIGHTS_ENGINE,
   MAX_ROWS,
   runReadOnlyQuery,
@@ -132,6 +133,30 @@ function isLogDestination(cfg: InsightConfig): boolean {
 }
 
 /** Engine label + table name for a destination, matching `readOnlyQuery.ts`. */
+/**
+ * The Redshift query target, schema-qualified.
+ *
+ * `redshiftSchema` is a documented setting that defaults to `public`, and core
+ * qualifies with it everywhere it builds Redshift SQL (`explorerSession.ts` in three
+ * places, `destinationTest.ts` in one). This package used the bare table name, which
+ * survived every test for a reason worth knowing: with the default `public` the
+ * unqualified name resolves through `search_path` and behaves identically. Set
+ * DURABLE_INSIGHT_REDSHIFT_SCHEMA to anything else and the same SQL silently reads a
+ * DIFFERENT table, or fails.
+ *
+ * Both the generated SQL and the guidance `describe_schema` hands the agent go
+ * through here, because a target the agent is TAUGHT that differs from the target
+ * queried is its own bug: the agent would write correct-looking SQL against a table
+ * that is not the one being read.
+ *
+ * Only Redshift is qualified this way, matching core exactly: Aurora has no schema
+ * setting, Athena's database is passed to the runner as a separate parameter, and
+ * DynamoDB and OpenSearch have no schema concept.
+ */
+function redshiftTarget(cfg: InsightConfig): string {
+  return `${cfg.redshiftSchema}.${cfg.redshiftTable}`;
+}
+
 function engineAndTable(cfg: InsightConfig): { engine: string; table: string } {
   switch (cfg.destinationType) {
     case "dynamodb":
@@ -141,7 +166,7 @@ function engineAndTable(cfg: InsightConfig): { engine: string; table: string } {
     case "aurora":
       return { engine: "PostgreSQL", table: cfg.auroraTable };
     case "redshift":
-      return { engine: "Redshift SQL", table: cfg.redshiftTable };
+      return { engine: "Redshift SQL", table: redshiftTarget(cfg) };
     case "opensearch":
       // The OpenSearch "table" is the index; buildSystemPrompt wraps it in
       // backticks in the FROM clause it generates.
@@ -357,7 +382,7 @@ function dialectFor(cfg: InsightConfig): Dialect {
     case "redshift":
       return {
         engine: "Redshift SQL",
-        from: cfg.redshiftTable,
+        from: redshiftTarget(cfg),
         idColumn: "execution_arn",
         listSelect:
           "execution_arn, status, function_name, start_time, end_time, duration_ms",
@@ -552,12 +577,65 @@ export async function runGetExecution(
 
 // ── list_executions ──────────────────────────────────────────────────────────
 
+/**
+ * Resolve the Logs Insights API window for a `list_executions` call.
+ *
+ * PRECEDENCE, and why:
+ *   1. An explicit `lookbackHours` wins -- the caller stated the window outright.
+ *   2. Otherwise `since` derives it, because a lower bound the caller asked to filter
+ *      on is worthless if the API never scanned that far back.
+ *   3. Otherwise the 24-hour default, matching `query`.
+ *
+ * The derived window is rounded UP to the next whole hour and given a small margin,
+ * so a `since` of "exactly 7 days ago" cannot land a fraction of a second inside the
+ * boundary and clip the oldest matching event.
+ *
+ * An unparseable `since` falls back to the default rather than throwing, which keeps
+ * this helper total. Note that this branch is UNREACHABLE through the tool: the query
+ * is built first, and `validateTimestamp` rejects a malformed bound there, so a bad
+ * value never arrives here. `tools.test.ts` pins that ordering, since it is the
+ * difference between a clean rejection and a silent 24-hour window.
+ */
+function logWindowFor(params: ListExecutionsParams): {
+  timeRangeMs: number;
+  lookbackHours: number;
+} {
+  const fromHours = (hours: number) => ({
+    timeRangeMs: hours * 60 * 60 * 1000,
+    lookbackHours: hours,
+  });
+
+  if (params.lookbackHours !== undefined && params.lookbackHours > 0) {
+    return fromHours(params.lookbackHours);
+  }
+
+  if (params.since !== undefined) {
+    const sinceMs = Date.parse(params.since);
+    if (!Number.isNaN(sinceMs)) {
+      const spanMs = Date.now() - sinceMs;
+      if (spanMs > 0) {
+        // +1 hour of margin, then round up: never scan less than asked for.
+        const hours = Math.ceil(spanMs / (60 * 60 * 1000)) + 1;
+        return fromHours(hours);
+      }
+    }
+  }
+
+  return fromHours(DEFAULT_LOG_TIME_RANGE_MS / (60 * 60 * 1000));
+}
+
 export interface ListExecutionsParams {
   status?: string;
   since?: string;
   until?: string;
   functionName?: string;
   limit?: number;
+  /**
+   * Log destinations only: how many hours back `StartQuery` should scan. Normally
+   * unnecessary -- when `since` is given the window is derived from it -- but needed
+   * to widen a search that has no lower bound.
+   */
+  lookbackHours?: number;
 }
 
 /**
@@ -734,6 +812,15 @@ export interface ListExecutionsResult {
   rows: string[][];
   count: number;
   truncated: boolean;
+  /**
+   * The window `StartQuery` actually scanned, in hours, for log destinations.
+   *
+   * Reported because a Logs Insights result is bounded by TWO things that can
+   * disagree: the API window, and the `filter` stages inside the query. If the window
+   * is narrower than the filters ask for, the result is silently partial -- so the
+   * window has to be visible to be trusted.
+   */
+  searchedLookbackHours?: number;
 }
 
 /**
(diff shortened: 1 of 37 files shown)


packages/durable-insight-mcp/src/tools.ts
```diff
@@ -0,0 +1,769 @@
...
+  // destination-appropriate bound (`ensureLimit` for logs, the row cap for SQL).
+  const query = isLogDestination(cfg)
+    ? buildListExecutionsLogsQuery(cfg, params)
+    : buildListExecutionsSql(cfg, params);
+  // Also pass the caller's limit as the row cap, NOT only as query text. For most
+  // destinations the generated `LIMIT`/`limit` already bounds the result, but
+  // DynamoDB PartiQL supports neither ORDER BY nor LIMIT, so its statement carries
+  // no bound at all -- without this, `limit: 10` on DynamoDB returns up to MAX_ROWS
+  // rows, which is the token cost this tool exists to avoid.
+  const result = await runReadOnlyQuery(cfg, query, {
```
yaythomas, line 858: `since`/`until` only become filters inside the query string, while this call passes no `timeRangeMs`; `runLogsInsightsReadOnly` therefore scopes `StartQuery` to its 24-hour default. For example, asking for executions since a week ago silently scans only the last day and can return incomplete results with `truncated: false`. 

If you derived the API window from the supplied bounds or expose/pass `lookbackHours`, and report the window that was searched.

packages/durable-insight-mcp/src/tools.ts
```diff
@@ -0,0 +1,769 @@
...
+        statusColumn: "status",
+        functionNameColumn: "function_name",
+        startTimeColumn: "start_time",
+        recordTypeColumn: undefined,
+        orderByAndLimit: true,
+      };
+    case "redshift":
+      return {
+        engine: "Redshift SQL",
+        from: cfg.redshiftTable,
```
yaythomas: With `DURABLE_INSIGHT_REDSHIFT_SCHEMA=analytics`, this currently generates `FROM workflow_insight`, so `get_execution`/`list_executions` can fail or query a different table through `search_path`. `engineAndTable` above needs the same qualification so `describe_schema` teaches the correct target too.

So you could include the configured Redshift schema in this target, e.g. `${cfg.redshiftSchema}.${cfg.redshiftTable}`. `redshiftSchema` is a documented setting, and the existing core Redshift paths schema-qualify the same table.

yaythomas (commented):
Two inline findings from review.

yaythomas (commented):
very nice :-)

packages/durable-insight-mcp/src/tools.ts
ParidelPooya: Addressed

packages/durable-insight-mcp/src/tools.ts
ParidelPooya: Addressed

yaythomas (approved):

PR #813, opened 2026-08-10
fix(insight): replace the JSON-extraction ReDoS and guard the whole class

ParidelPooya:
CodeQL alert **#247** (`js/polynomial-redos`, **high**) on merged main, against
  `packages/durable-insight-core/src/verdict.ts:30`:

  ```js
  text.match(/\{[\s\S]*"satisfied"[\s\S]*\}/)

  Two unbounded [\s\S]* around a literal is polynomial. On input that starts a plausible
  object and never finishes one, the engine retries the inner scan from every start
  position. Measured at 4× per doubling:

  ┌───────┬───────┐
  │ input │ time  │
  ├───────┼───────┤
  │ 2k    │ 6 ms  │
  ├───────┼───────┤
  │ 8k    │ 96 ms │
  ├───────┼───────┤
  │ 32k   │ 1.5 s │
  ├───────┼───────┤
  │ ~128k │ ~24 s │
  └───────┴───────┘

  The input is a model reply, so its length is not ours to bound, and a verbose reply
  reaches those sizes easily.

  Sweeping the class found three more

  None of them flagged, all measured quadratic:

  ┌───────────────┬─────────────────────────────────┬────────────────────────┬─────────────────┐
  │ Location      │ Pattern                         │ Input                  │ Exploitable     │
  ├───────────────┼─────────────────────────────────┼────────────────────────┼─────────────────┤
  │ verdict.ts:30 │ /\{[\s\S]*"satisfied"[\s\S]*\}/ │ model reply            │ yes — the alert │
  ├───────────────┼─────────────────────────────────┼────────────────────────┼─────────────────┤
  │ llm.ts:600    │ /\{[\s\S]*"query"[\s\S]*\}/     │ model reply (Copilot)  │ yes             │
  ├───────────────┼─────────────────────────────────┼────────────────────────┼─────────────────┤
  │ llm.ts:827    │ /\{[\s\S]*"query"[\s\S]*\}/     │ model reply (local)    │ yes             │
  ├───────────────┼─────────────────────────────────┼────────────────────────┼─────────────────┤
  │ llm.ts:727    │ /\s*\(.*\)$/                    │ hardcoded preset label │ no              │
  └───────────────┴─────────────────────────────────┴────────────────────────┴─────────────────┘

  The fourth is fixed anyway — one line — but it is not a vulnerability and CodeQL was
  right not to report it. Its input is one of three compile-time constants; the shape is
  removed so that making labels dynamic later cannot quietly reintroduce a slow path.

  I also measured nine other candidates in this package and they are linear. Worth
  stating so this change isn't read as having missed them: queryShape.ts:113/178/179/302,
  agentLoop.ts:370, and the four hostModuleScan.ts import matchers. Their quantifiers
  cannot compete — disjoint character classes, a literal pinned between them, a lazy
  quantifier, or anchors at both ends. These are the nine allowlisted below.

  What the old regex actually meant

  Worth writing down, because it was never a JSON parser and never balanced braces. With
  both quantifiers greedy and matching leftmost-longest, it always resolved to "first {
  to last }, marker somewhere between". extractJsonObject computes that span by
  index, which is linear.

  Equivalence is asserted against the reconstructed regex over sixteen hand-picked cases
  plus a 20,000-input fuzz over a brace-heavy alphabet, because a performance fix that
  changes behavior is just a different bug.

  That check paid for itself immediately. stripTrailingParenthetical used
  lastIndexOf("("), which looks obviously right and is wrong: a regex match is leftmost,
  so \s*\(.*\)$ anchors on the earliest ( that still has the final ) after it.
  "Trailing (a) (b)" → "Trailing", not "Trailing (a)". Only the differential test
  caught it.

  The real fix is the guard, not the four edits

  This is the third round of this finding. #809 fixed three trailing-anchor regexes
  after CodeQL flagged one; then CodeQL flagged a fourth of a different shape, and sweeping
  for that shape turned up three more.

  Each round fixed instances. The recurring mistake was scoping the sweep to the shape just
  found, so the next shape stayed invisible — the same mistake as guarding vscode and
  missing electron (#809), and guarding run*Query and missing fetchAthenaRecord
  (#810). So redosGuard.test.ts works on the class, with three rules:

  - R1 — two or more wide quantifiers, which can compete for the same characters.

    The JSON-extraction shape.

  - R2 — any quantifier, wide or a single escaped literal, with a trailing $ and no

    leading ^. When the anchor is unreachable the engine retries from every start
    position. This is the #809 shape, and it needs no wide class at all. A leading ^
    removes the risk — one place to begin.

  - R3 — a repeated group. (a+)+ is exponential rather than polynomial, and (ab)+

    can backtrack across alternatives. None exist here today; the rule is cheap and the
    failure mode is the worst of the three.

  Anything matching a rule must be allowlisted with a measured justification. Nine
  patterns are, each timed at 2k and 8k characters of adversarial input rather than reasoned
  about — quadratic shows as ~16×, linear as ~1×. The failure message names file, line,
  pattern and matched rule, because the rule tells you which adversarial input to build.

  An allowlist rather than a ban because these shapes are necessary but not sufficient
  conditions, and a guard that cries wolf gets deleted. Adding an entry is the point at
  which to measure: if you cannot state the pathological input and show it is not, the
  pattern does not belong there.

  The guard was itself written to the narrower spec

  This is the substance of the second and third commits, and it is in the file's header
  rather than only here, because a PR description does not travel with the code.

  The first version implemented only R1, so it scored both regexes #809 removed below
  threshold — /\/+$/ at literally zero, since an escaped literal is not a wide class —
  and (a+)+$, the textbook exponential case, at zero as well. A file arguing at length
  against narrow sweeps was itself scoped to the shape just found. Widening it cost
  nothing in churn: the same nine patterns are found as before, so R2 and R3 earn their
  place entirely against the future.

  The more important addition is the detector self-tests. The sweep assertions prove the
  current sources are clean, which says nothing about whether the detector can find
  anything — they pass just as well against a detector narrowed to uselessness, which is
  exactly how v1 shipped. riskRules is now tested directly against every shape this
  repository has had to remove, plus two canonical exponential cases, plus three benign
  patterns it must not flag. Reverting to the R1-only spec fails five named tests.

  The third commit brings the header into line: four places still described the R1-only rule,
  so a reader formed that model and then met R2 and R3 seventy lines down in riskRules.
  That mismatch is the exact failure the widening existed to fix.

  Verification

  Mutation-verified in both directions:

  ┌───────────────────────────────────┬───────────────────────────────────────────────────┐
  │ Mutation                          │ Result                                            │
  ├───────────────────────────────────┼───────────────────────────────────────────────────┤
  │ Reintroduce the extraction regex  │ pathological test takes 58 s, fails the 1 s bound │
  ├───────────────────────────────────┼───────────────────────────────────────────────────┤
  │ Swap trimEnd() for /\s+$/         │ 65 s, fails                                       │
  ├───────────────────────────────────┼───────────────────────────────────────────────────┤
  │ Add an unseen shape /<a.*href.*>/ │ reported by file, line and rule                   │
  ├───────────────────────────────────┼───────────────────────────────────────────────────┤
  │ Revert to the R1-only detector    │ 5 named self-tests fail                           │
  ├───────────────────────────────────┼───────────────────────────────────────────────────┤
  │ Neuter the sweep to find nothing  │ non-vacuity assertions fail                       │
  └───────────────────────────────────┴───────────────────────────────────────────────────┘

  Timing bounds are deliberately loose — 1 s against a measured ~2 ms — so they cannot flake
  on a slow runner while still failing a quadratic implementation by orders of magnitude.

  Tests 336 → 380 in core. durable-insight-mcp (404), -vscode (33) and -desktop
  (48) are untouched. Lint, typecheck:hosts, renderer typecheck and both host bundles
  green.

  Notes for the reviewer

  - Three things I got wrong, all found by mutation or by review rather than by reading.

    My first timing test used "name" + " ".repeat(n) + "(x)", which does not catch
    /\s+$/ — after slicing at the parenthesis the prefix ends in whitespace, so the
    anchor succeeds on its first viable start and the scan is linear; the quadratic case
    needs the run followed by a non-space. The guard flagged jsonExtract.ts for
    documenting the pattern it replaced, the fourth time a scanner here has reported its
    own prose, so the sweep reads code lines only. And a comment claimed the failure message
    named the matched rule when it did not — fixed by making it true.

  - One intended divergence. A parenthetical containing a newline is stripped by

    stripTrailingParenthetical where the regex left it, because . does not match a line
    terminator. Unreachable today — the only inputs are three single-line preset labels —
    and stripping is what the caller wants. Documented as the contract and pinned by a test
    asserting both behaviors, rather than sitting silently next to an equivalence claim
    that does not hold for it.

  - CodeQL remains the primary detector. This guard is a lexical scan: it can miss a

    literal spelled unusually and does not see new RegExp built from variables. It is a
    floor on coverage, not a proof of absence — its job is stopping a known class from
    returning between scans, and forcing the measurement conversation when someone adds one.

  - The MCP package has zero patterns in this class today, so the guard stays in core

    rather than expanding this change. Generalizing it — the mirror-guard shape already used
    for hostAgnostic.test.ts — is a reasonable follow-up.

  - Alert #247 should close on the next scan.
  - The three commits are worth reading in order if you want the reasoning; the squash will

    flatten them.

  Testing (per CONTRIBUTING)

  Unit tests for both helpers, differential equivalence against the regexes they replace,
  timing bounds on pathological input, and the class guard with its own self-tests. Every
  guard mutation-verified. No conformance tests added. Not an SDK feature or API change, so
  no examples needed.

### DIFF
--- a/packages/durable-insight-core/src/redosGuard.test.ts
+++ b/packages/durable-insight-core/src/redosGuard.test.ts
@@ -1,0 +1,204 @@
+/**
+ * Class-level guard against polynomial-ReDoS regexes in this package.
+ *
+ * WHY THIS EXISTS:
+ * This is the third round of the same finding. #809 fixed three trailing-anchor
+ * regexes (`/\/+$/`, `/\s+$/`) after CodeQL flagged one of them. CodeQL then flagged
+ * a fourth in `verdict.ts` — a different shape, two unbounded `[\s\S]*` around a
+ * literal — and sweeping for THAT shape turned up three more the scanner had not
+ * reported.
+ *
+ * Each round fixed instances. The recurring mistake was scoping the sweep to the
+ * shape that had just been found, so the next shape was invisible: exactly the
+ * mistake made with `HOST_MODULES` (guarding `vscode` and missing `electron`) and with
+ * the query choke point (guarding `run*Query` and missing `fetchAthenaRecord`). So
+ * this guard is about the CLASS: any regex with two or more unbounded quantifiers must
+ * be justified, whatever its shape.
+ *
+ * HOW IT WORKS:
+ * Every regex literal in this package's non-test sources is extracted and scored by
+ * how many unbounded quantifiers it contains (`.*`, `.+`, `[^x]*`, `\s+`, ...). Two or
+ * more means the engine can be made to retry an inner scan from many start positions,
+ * which is the necessary condition for polynomial backtracking. Such a pattern must
+ * appear in ALLOWED below, with a note on why it is safe.
+ *
+ * WHY AN ALLOWLIST RATHER THAN A BAN:
+ * Two unbounded quantifiers are necessary but not sufficient — seven patterns here are
+ * genuinely linear because their quantifiers cannot overlap (different character
+ * classes, or a literal pinned between them). Every entry was MEASURED, not reasoned
+ * about: growth from 2k to 8k characters of adversarial input, where quadratic shows
+ * as ~16x and linear as ~1x. Banning the shape outright would force pointless rewrites
+ * of correct code, and a guard that cries wolf gets deleted.
+ *
+ * ADDING AN ENTRY IS THE POINT AT WHICH TO MEASURE. If you cannot state the input that
+ * would be pathological and show it is not, the pattern does not belong here.
+ *
+ * Test files are excluded from the scan on purpose: this file's own ALLOWLIST contains
+ * pattern text, and two earlier scanners in this repo flagged their own documentation.
+ */
+import { readFileSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+
+const SRC = __dirname;
+
+/**
+ * Patterns with two or more unbounded quantifiers that are nonetheless linear.
+ * Keyed by the regex SOURCE rather than by file and line, so ordinary edits do not
+ * invalidate the list.
+ */
+const ALLOWED = new Map<string, string>([
+  [
+    String.raw`\bfrom\s*["'](\.[^"']*)["']`,
+    "hostModuleScan: `\\s*` matches only whitespace and `[^\"']*` stops at a quote, " +
+      "so the two cannot compete for the same characters. Measured flat from 2k to 8k.",
+  ],
+  [
+    String.raw`(?:^|;)\s*import\s*["'](\.[^"']*)["']`,
+    "hostModuleScan: as above — disjoint character classes separated by literals.",
+  ],
+  [
+    String.raw`\brequire\s*\(\s*["'](\.[^"']*)["']\s*\)`,
+    "hostModuleScan: as above.",
+  ],
+  [
+    String.raw`\bimport\s*\(\s*["'](\.[^"']*)["']\s*\)`,
+    "hostModuleScan: as above.",
+  ],
+  [
+    String.raw`\blimit\s+\d+`,
+    "agentLoop: `\\s+` and `\\d+` are disjoint, so neither can absorb the other's " +
+      "characters. Measured flat.",
+  ],
+  [
+    String.raw`\|\s*fields\s+([^|]+)`,
+    "queryShape: `[^|]+` runs to the next pipe or end of input, and the `fields` " +
+      "literal pins the start. Measured flat.",
+  ],
+  [
+    String.raw`^\s*\*\s*$`,
+    "queryShape: anchored at both ends with a literal `*` between the quantifiers, " +
+      "so there is one candidate split. Measured flat.",
+  ],
+  [
+    String.raw`\*\s*,|\bselect\s+\*`,
+    "queryShape: an alternation of two short anchored branches, not nested " +
+      "quantifiers. Measured flat.",
+  ],
+  [
+    String.raw`^([\s\S]*?)(\s+from\s)`,
+    "queryShape: start-anchored, and the first quantifier is LAZY, so it extends one " +
+      "character at a time toward a single required literal rather than backtracking " +
+      "over the whole input. Measured flat.",
+  ],
+]);
+
+/** Every non-test `.ts` file in this package, recursively. */
+function sourceFiles(dir: string): string[] {
+  const out: string[] = [];
+  for (const entry of readdirSync(dir, { withFileTypes: true })) {
+    const full = join(dir, entry.name);
+    if (entry.isDirectory()) out.push(...sourceFiles(full));
+    else if (entry.name.endsWith(".ts") && !entry.name.endsWith(".test.ts")) {
+      out.push(full);
+    }
+  }
+  return out;
+}
+
+interface Found {
+  file: string;
+  line: number;
+  source: string;
+  unbounded: number;
+}
+
+/**
+ * Extract regex literals and count unbounded quantifiers.
+ *
+ * Deliberately a lexical scan rather than a parse. It can miss a literal spelled in an
+ * unusual way, and it does not see `new RegExp(...)` built from variables — so it is a
+ * floor on coverage, not a proof of absence. CodeQL remains the primary detector; this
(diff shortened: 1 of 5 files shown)


packages/durable-insight-core/src/redosGuard.test.ts
```diff
@@ -0,0 +1,388 @@
...
+    // A stale entry silently widens the rule. If a pattern is gone, its exemption
+    // should go with it.
+    const sources = new Set(risky.map((r) => r.source));
+    const orphans = [...ALLOWED.keys()].filter((k) => !sources.has(k));
+    expect(orphans).toEqual([]);
+  });
+
+  it("every risky pattern is allowlisted with a reason", () => {
+    const unjustified = risky
+      .filter((r) => !ALLOWED.has(r.source))
```
yaythomas: ALLOWED is checked only by regex source, so the known-vulnerable \s*\(.*\)$ exemption intended for jsonExtract.test.ts also exempts the same regex in production code.

mutation-added a vulnerable pattern to verdict.ts, and the guard still passed.

could scope fixture-only exemptions by file plus source, or otherwise enforce the location recorded in the justification?

packages/durable-insight-core/src/redosGuard.test.ts
```diff
@@ -0,0 +1,388 @@
...
+  // R1 -- two or more wide quantifiers can compete for the same characters, which is
+  // the JSON-extraction shape (`[\s\S]*` ... `[\s\S]*`).
+  const wide = [...source.matchAll(WIDE_QUANTIFIER)].length;
+  if (wide >= 2) found.push(`R1: ${wide} wide quantifiers`);
+
+  // R2 -- one quantifier plus an END anchor and no START anchor. When the anchor is
+  // unreachable the engine retries from every position, which is exactly the shape
+  // CodeQL flagged in #809 (`/\/+$/`, `/\s+$/`). A start anchor removes it: there is
+  // then only one place to begin.
+  const anyQuantifier = [...source.matchAll(ANY_QUANTIFIER)].length;
```
yaythomas: The whole-pattern startsWith("^")/endsWith("$") checks assume those anchors govern every branch. Both /a+$|x/ and /^x|a+$/ score zero even though their a+$ branch is quadratic. Mutation tests failed as expected, and timings grew approximately 3 → 12 → 52 → 208 ms from 2k to 16k characters.

Could add self-tests and make R2 alternation-aware, or conservatively flag the shape.

ParidelPooya: Fixed

packages/durable-insight-core/src/redosGuard.test.ts
ParidelPooya: Fixed

packages/durable-insight-core/src/redosGuard.test.ts
```diff
@@ -0,0 +1,518 @@
...
+      current += char;
+      if (char === "]") inClass = false;
+      continue;
+    }
+    if (char === "[") {
+      inClass = true;
+      current += char;
+      continue;
+    }
+    if (char === "|" || char === "(" || char === ")") {
```
yaythomas, line 273: Splitting at every parenthesis discards the relationship between a quantified atom and its branch’s trailing anchor. For example, /^x|(?:a+)$/ becomes separate ?:a+ and $ fragments, while the whole expression is ignored because it starts with ^. R2 therefore returns zero findings. The same occurs with /^x|(?:a+)b$/.

Both patterns measured quadratic growth: approximately 3, 12, 52, and 208 ms from 2k through 16k characters. I also planted /^x|(?:a+)$/ in verdict.ts; the complete guard remained green.

Might be an idea to preserve grouping context when identifying branches and add this case as a regression test.

yaythomas (approved):

PR #817, opened 2026-08-10
feat(sdk): support runtimes without AsyncLocalStorage or util.formatWithOptions

ParidelPooya:
LLRT and other lightweight JavaScript runtimes implement a subset of the Node API. Two of the gaps land directly on this SDK, and both are fatal rather than degrading: `async_hooks` has no `AsyncLocalStorage`, so importing `withDurableExecution` throws at module scope, and `node:util` has no `formatWithOptions`, so the default logger throws on the first multi-argument log record. Durable functions support container images, which makes a runtime the service does not manage a supported way to deploy -- the account used for testing already ran Go durable functions packaged exactly that way.

Feature-detect both. `createContextStorage()` prefers the runtime's own `AsyncLocalStorage` and otherwise returns a fallback, and the logger falls back to `util.format`. On a supported managed runtime neither fallback is reachable (`AsyncLocalStorage` exists in every Node >= 12.17 and this package requires Node >= 22), so the change is a no-op there.

A faithful `AsyncLocalStorage` polyfill is not possible on these runtimes: they emit `init` for promise resources but no `before`/`after`, so there is no callback in which to reinstate a store when a continuation runs, and `await` on a native promise never calls a patched `then`. The fallback is therefore correct-or-silent -- the store is live for the synchronous portion of `run()`, including nested frames entered synchronously, and reports `undefined` after the callback suspends. Never reporting a *stale* store is the point: `validateContextUsage` terminates the execution when the active context disagrees with the operation's, so a stale value would fail correct workflows, whereas `undefined` makes the check skip.

Checkpointing and replay are unaffected. What degrades is observability, and the SDK now says so once per execution environment rather than leaving it to be inferred from log records missing their operation metadata:

- log records emitted after an `await` lose operationId/operationName/attempt
- replayed log records are no longer suppressed, so a log line behind the replay frontier is re-emitted once per replay
- using a parent or sibling context inside `runInChildContext` is only detected when it happens synchronously; `no-nested-durable-operations` in `@aws/durable-execution-sdk-js-eslint-plugin` catches the rest at build time

Tests cover the fallbacks from two directions, both in the existing suite, so the pull request build gains no new dependencies and no new job:

- unit tests for each fallback in isolation, including the absent-capability path
- `constrained-runtime.composed.test.ts`, which runs the real engine -- steps, a retry, a suspending wait, a child context, a parallel block -- twice in one process: once normally, once with `async_hooks` and `node:util` mocked down to what a constrained runtime provides. It asserts the same result, the same invocation count, the same operation transitions and byte-identical checkpoint requests, which is what pins the fallbacks to observability and nothing else. The service is stood in for in memory, because the testing SDK's runner needs `node:worker_threads` and fake timers.

`llrt-container-tests.yml` covers the part that cannot be simulated -- the deployment shape -- by building a container image around the LLRT binary and running a durable function on real Lambda. It is dispatch-only, since it pushes an image to ECR and creates a function.

Verified: 1176 unit tests (18 added), tsc, eslint unchanged from the base commit (20 warnings, 0 errors), the ESM, CJS and runtime-detection integration tests, and an api-extractor report that regenerates byte-identically -- the new module is internal, so the public API surface does not move. Checkpoint wire traffic is byte-identical to the published 2.2.0, so upgrading cannot disturb an in-flight execution. The container workflow was run end to end against real Lambda (us-east-1, arm64, LLRT v0.8.1-beta): the execution succeeded across 3 invocations and 10 operations with the same result as a Node run, at 33 MB peak memory.


yaythomas (approved):

yaythomas:
Files: .github/workflows/llrt-container-tests.yml:71, .github/workflows/scripts/llrt/llrt-cloud-test.js:111

The workflow passes LAMBDA_ENDPOINT, but this client is constructed with only { region: REGION }. Unlike AWS_ENDPOINT_URL_LAMBDA, the SDK does not automatically consume LAMBDA_ENDPOINT. Our existing integration driver explicitly assigns it to clientConfig.endpoint.

If the test account requires this endpoint, this workflow will call the public endpoint instead. Could we either wire process.env.LAMBDA_ENDPOINT into the client or remove the unused workflow variable if public Lambda is intentional?

2. Each dispatch leaves an orphaned log group
llrt-cloud-test.js:402



cleanUp() deletes the function and ECR image, but Lambda does not delete /aws/lambda/${FUNCTION_NAME} when the function is deleted. Because every dispatch uses a unique GITHUB_RUN_ID, this leaves one permanent log group per run. The existing integration driver uses DeleteLogGroupCommand; could this cleanup do the same, preferably best-effort

PR #827, opened 2026-08-13
perf(eslint-plugin): replace redundant AST walks with ESLint scope analysis

ParidelPooya:
`no-closure-in-durable-operations` and `no-non-deterministic-outside-step` both re-traversed subtrees that ESLint's visitor had already walked, redundantly enough to be super-linear in file size. Neither rule re-parsed source, but neither used the scope information ESLint derives while parsing.

**`no-closure-in-durable-operations`** recomputed `getVariablesDeclaredInScope()` and `findOuterScopeVariables()` for *every assignment it found* — each a full walk of the enclosing callback — on top of an outer walk of the same subtree. It now reads scope data instead: for a durable callback's function scope, `scope.through` is exactly the set of references that could not be resolved inside the callback, so filtering it to writes yields the violations with no traversal at all. `walkNode`, `walkForDeclarations`, `getVariablesDeclaredInScope`, `findOuterScopeVariables` and `isAssignment` are gone; the rule is a third of its previous size.

**`no-non-deterministic-outside-step`** walked every ancestor of every `CallExpression`, `NewExpression` and `MemberExpression` to answer "am I inside a step?", and resolved transitive non-determinism with a `Program:exit` fixpoint loop that re-walked every function body on each pass. The ancestor walk is now an O(1) `stepDepth` counter maintained by the enter/exit visitors, and the fixpoint loop is a caller graph built during the traversal ESLint already performs, propagated with a worklist.

That graph is keyed on the function nodes callees resolve to, not on identifier names. Names would let a non-deterministic function taint every call to that name in the file:

```js
function moduleA() { function h() { Date.now(); } h(); }
function moduleB() { function h() { return 1; } return h(); }  // must not be reported
```

Only the innermost enclosing function is recorded per call site; outer functions are reached through a nesting map during propagation. Recording all of them is O(calls × nesting depth), and an earlier iteration that did so measured *slower* than the code it replaced on deeply nested files.

The old walkers also enumerated non-AST properties via `for...in`, recursing into `range`, `loc`, `start` and `end` as if they were nodes. On a 32-node file the walker visited 462 objects — 14.4x the real node count.

#### Correctness fixes

Using real scope data fixes four latent bugs, each covered by a test:

- Shadowed variables in nested functions no longer produce false positives.
- Destructured declarations (`let { total } = event`) are seen.
- Variables declared in outer nested blocks are no longer missed.
- A named function expression assigning to its own name is no longer reported — that binding lives in a wrapper scope outside the function scope, but it is not outer state.

#### ⚠️ Behavior changes

Four surface **new** errors for upgrading users:

1. **Module-scope variables are reported.** `findOuterScopeVariables` only walked enclosing *function* nodes, so a top-level `let counter = 0`, `export let count`, or a `catch` binding mutated inside a step produced nothing. Module-level mutable state around a handler is a common pattern, so this is likely the highest-volume new error in real codebases.
2. **Destructuring and for-of assignment targets are caught**: `[a] = [1]`, `({ a } = obj)`, `for (a of xs)` were all missed by the old `isAssignment` check.
3. **Calls to functions declared after the call site are reported.** `callsToCheck` was filtered against a partially built map during traversal, so hoisted declarations were skipped.
4. A function is treated as non-deterministic if any step it contains uses a non-deterministic operation — already true, but unreachable for hoisted declarations.

One **removes** reports: assigning a function to a member (`obj.method = function () { Date.now(); }`) registered it under the property name, so a call to an unrelated bare `method()` was reported. A bare identifier call never refers to a member assignment.

`no-nested-durable-operations` is unchanged — its only ancestor walk is O(depth) per durable-operation call.

### Demo/Screenshots

Best of 5, ESLint 9.39.4, measured against the parent commit. **Multipliers here track the ratio of rule work to file size**, so they compress as fixtures grow and parse time dominates both sides. Fixtures are described so the numbers can be reproduced.

**N assignments to an outer variable inside one step callback:**

| N | before | after |
|---|---|---|
| 100 | 19.1 ms | 1.4 ms |
| 200 | 68.9 ms | 2.1 ms |
| 400 | 263.9 ms | 1.8 ms |

The new rule is flat in N, so no single multiplier is meaningful. Growth per doubling went from ~3.5x (quadratic) to none.

**Chain of N functions each calling the next, declared in dependency order:**

| N | before | after |
|---|---|---|
| 40 | 2.3 ms | 2.0 ms |
| 80 | 3.3 ms | 3.1 ms |
| 160 | 6.0 ms | 6.1 ms |

No meaningful difference — the old fixpoint converged in about two passes. Included because it is the honest common case.

**The same chain declared in reverse order, forcing one pass per function:**

| N | before | after |
|---|---|---|
| 40 | 7.7 ms | 1.6 ms |
| 80 | 27.8 ms | 2.7 ms |
| 160 | 97.7 ms | 6.5 ms |

Not like-for-like: the new rule emits 43/83/163 messages against 3 (the hoisted-declaration fix) and is still faster.

**Nested functions to depth D, each level declaring the next and making 20 calls:**

| D | before | after | parse only |
|---|---|---|---|
| 50 | 4.8 ms | 4.4 ms | 4.2 ms |
| 100 | 13.4 ms | 11.2 ms | 11.0 ms |
| 200 | 41.8 ms | 36.0 ms | 27.6 ms |

Both sides are close to parse cost on this shape. This is the fixture where recording every enclosing function per call site measured 7.9 / 19.3 / 65.6 ms — slower than before the change, which is why the innermost-only design exists.

Coverage of both rewritten rules:

```
File                                   | % Stmts | % Branch | % Funcs | % Lines
no-closure-in-durable-operations.ts    |     100 |      100 |     100 |     100
no-non-deterministic-outside-step.ts   |     100 |      100 |     100 |     100
```

### Testing

Yes. 90 tests pass (up from 63), with both rewritten rules at 100% statement, branch, function and line coverage.

The existing `.composed.test.ts` RuleTester suites pass **unmodified**, which is the main evidence that behavior is preserved — including the deliberate double-reporting of `Date.now()` (CallExpression + MemberExpression) and the nested-context duplicate report.

The mock-AST unit tests for `no-closure-in-durable-operations` had to be rewritten against `RuleTester`, because hand-built AST fixtures carry no scope information. The only remaining mocks cover the ESLint version fallback and the no-scope-information path, which `RuleTester` cannot reach.

New tests cover: same-named functions in sibling scopes, hoisted declarations, arrow functions bound to variables, unresolvable callees (imported helpers must not be reported), namespaced `uuid.v4()`, module-scope and `export let` mutations, `catch` bindings, array/object destructuring and for-of assignment targets, shadowing in nested functions, a named function expression assigning to its own name, recursion through a named function expression, and only the first function argument of a durable operation being checked.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-eslint-plugin/src/rules/no-closure-in-durable-operations/no-closure-in-durable-operations.test.ts
+++ b/packages/aws-durable-execution-sdk-js-eslint-plugin/src/rules/no-closure-in-durable-operations/no-closure-in-durable-operations.test.ts
@@ -1,3 +1,4 @@
+import { RuleTester } from "eslint";
 import { noClosureInDurableOperations } from "./no-closure-in-durable-operations";
 
 describe("no-closure-in-durable-operations", () => {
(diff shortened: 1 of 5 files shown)


yaythomas (approved):
:shipit:

I walked through both rule rewrites and the scope/call-graph semantics.
The use of `scope.through` correctly identifies closure writes, and the
function-node-based propagation avoids the same-name collisions from the
previous implementation.

I also verified the affected package locally: all 90 tests and the build
pass, both rewritten rules have 100% coverage, and smoke tests pass with
ESLint 8, 9, and 10. I couldn't find any actionable regressions.

Issue #831, opened 2026-08-14
[otel] End all recording spans on non-terminal invocations


zhongkechen:
## Summary

The JavaScript OTel plugins intentionally drop references to recording spans without ending them when a durable execution returns a non-terminal status.

Affected behavior:

- `ExecutionOtelPlugin` clears open operation and attempt spans during invocation cleanup without ending them.
- Both `ExecutionOtelPlugin` and `InvocationOtelPlugin` drop the execution-scoped `Workflow` span on `PENDING` or `RETRYING`.

OpenTelemetry spans are only complete and eligible for processing/export after `span.end()` is called.

## Evidence

- `packages/aws-durable-execution-sdk-js-otel/src/execution-plugin.ts`
  - `onInvocationEnd` explicitly discards open operation spans without ending them.
  - The non-terminal workflow path drops `workflowSpan` without ending it.
- `packages/aws-durable-execution-sdk-js-otel/src/invocation-plugin.ts`
  - The non-terminal workflow path drops `workflowSpan` without ending it.
- `execution-plugin-invocation-lifecycle.test.ts` explicitly asserts that non-terminal workflow spans are never ended or exported and clears unfinished attempt state.
- `invocation-plugin.test.ts` explicitly asserts that non-terminal workflow spans are never ended or exported.

## OpenTelemetry requirement

The Trace API defines `End` as the operation that completes a span and makes it ready for delivery through the configured span processor:

https://opentelemetry.io/docs/specs/otel/trace/api/#end

## Impact

- Span processors can observe span starts without corresponding ends.
- Processor state or references may remain live across warm Lambda invocations.
- Re-created deterministic spans can produce repeated starts for the same logical span ID.
- The behavior prevents a strict OpenTelemetry compliance claim.

## Acceptance criteria

- Every recording span created by either JavaScript OTel plugin is ended exactly once.
- Non-terminal workflow and suspended-operation behavior does not rely on abandoning a recording span.
- Cross-invocation workflow and operation correlation remains intact.
- Tests retain span references and assert that they are no longer recording after invocation cleanup.
- Tests cover `PENDING`, `RETRYING`, terminal success, and terminal failure for both plugin variants.

Related cross-SDK issue: aws/aws-durable-execution-sdk-python#642

ayushiahjolia:
https://github.com/aws/aws-durable-execution-sdk-js/pull/850

PR #839, opened 2026-08-17

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-js/issues/831

*Description of changes:*
Ends every recording span created by ExecutionOtelPlugin and InvocationOtelPlugin exactly once, including on PENDING/RETRYING (non-terminal) invocations.

Previously, the Workflow span (both plugins) and any open Operation/Attempt spans (suspended WAIT/CALLBACK, in-flight STEP attempts) were dropped without calling .end() when an invocation suspended rather than terminating. Spans that are never ended are never delivered to span processors/exporters, which breaks strict OTel compliance and can leak un-ended recording spans across warm Lambda invocations.

Fix: always end the Workflow span and any still-open Operation/Attempt spans in onInvocationEnd, regardless of terminal
status, using the same status-mapping already used for the Invocation span (PENDING -> OK, RETRYING -> UNSET). Cross-invocation correlation is unaffected since spans keep their deterministic IDs.

Tests updated to assert spans are ended/exported on PENDING/RETRYING instead of asserting they're dropped; added coverage for in-flight attempt spans left open by a suspending invocation.


ayushiahjolia:
https://github.com/aws/aws-durable-execution-sdk-js/pull/850

PR #850, opened 2026-08-21

ayushiahjolia:
*Issue #, if available:*
https://github.com/aws/aws-durable-execution-sdk-js/issues/831

*Description of changes:*
Both OTel plugins created a recording Workflow span at invocation start and dropped it un-ended on non-terminal (PENDING/RETRYING) invocations. An un-ended span is never delivered to span processors, so it leaked across warm Lambda invocations and was never exported. In-flight attempt spans (and, in ExecutionOtelPlugin, suspended operation spans) were also left un-ended.

Carry the Workflow span identity as a non-recording span context while the execution is in flight, and create the single recording Workflow span once, at the terminal invocation, backdated to the execution start. End in-flight attempt and suspended operation spans at invocation cleanup; a later invocation that resumes an operation exports it under its own ID rather than reusing the deterministic one, so no two exported spans share a (traceId, spanId). Resolve the non-recording context's sampling decision from the provider's own sampler (resolveWorkflowRoot) so operations are sampled consistently with the eventual root and enrichLogContext() reports the true sampling flag.

Also fixes a latent bug in ExecutionOtelPlugin where onOperationStart firing twice for one operation (internal retry) overwrote the span map and orphaned the first span.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **otel**: the OTel plugins no longer leave recording spans un-ended on non-terminal
+  (`PENDING`/`RETRYING`) invocations, and no longer export two spans sharing one
+  `(traceId, spanId)`. The Workflow span identity is now carried as a non-recording span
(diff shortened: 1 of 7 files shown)


ayushiahjolia:
https://github.com/aws/aws-durable-execution-sdk-js/pull/882

PR #856, opened 2026-08-24
chore: ship LICENSE/NOTICE in every released package and gate it in CI

ParidelPooya:
d7bbfc3 added the missing Apache-2.0 `license` field to the ESLint plugin after it had already published 0.0.1 through 1.1.0 without one. Nothing checked, so it took a human noticing. This adds that check, and fixes the rest of what the same audit turned up.

The `license` field is only a metadata pointer, so every published tarball named Apache-2.0 while carrying none of its text: no package had a LICENSE or NOTICE file, and none listed one in `files`. The Python SDK ships both in its wheels; the JS repo was the outlier. All six released packages now carry copies of the repo root files.

Three integration-test packages (cdk-bundling, esm, lambda-runtime-detection) had neither `private: true` nor a license, leaving them publishable by accident via `npm publish --workspaces`. Marked private. Also fills in the missing `license` on insight-core and cjs-integration-test, and restores the trailing newline d7bbfc3 dropped.

check-legal-files.js enforces four rules across the workspace: a declared license everywhere; every package either private or on the release list; and for released packages, LICENSE/NOTICE present, byte-identical to the root copies, and named in `files`. The release list is parsed out of iterate-publish-npm.sh rather than duplicated, so adding a package to the release list without legal files fails the check. It runs on every PR and again before publish, where a failure is still recoverable -- published versions are immutable.

Committed files rather than a build-time copy step: `files` cannot reference paths outside the package directory, so the npm equivalent of hatchling's force-include needs a prepack script, and anything generated at build time vanishes under `npm publish --ignore-scripts` or a publish from an unbuilt tree. That is the same silent-omission failure being fixed here. Verified with `npm pack --dry-run --ignore-scripts`: all six tarballs contain both files. The drift risk this creates is covered by the byte-comparison rule.


.github/workflows/scripts/legal/check-legal-files.js
```diff
@@ -0,0 +1,196 @@
...
+            `ships in the npm tarball.`,
+        );
+      } else if (!readFileSync(legalPath).equals(canonical.get(legalFile))) {
+        errors.push(
+          `${dir}/${legalFile}: does not match the repo root ${legalFile}`,
+        );
+      }
+
+      const files = manifest.files;
+      if (!Array.isArray(files)) {
```
yaythomas: should this be inside the LEGAL_FILES loop?

benign enough, since `assert.ok(errors.some(...))` will mask this, but you'll see a must declare a "files" array twice.

yaythomas (commented):
currently publish-workflow check runs after npm ci / build / test (several minutes in) despite being dependency-free. Moving it right after checkout would fail faster. The comment's guarantee ("before anything reaches the registry") holds either way, so this is purely ergonomic.

yaythomas (approved):

.github/workflows/scripts/legal/check-legal-files.js
ParidelPooya: Fixed

PR #859, opened 2026-08-25
fix(examples): pin Managed Instances runtime to unblock capacity provider tests

ParidelPooya:
## Summary

`capacity-provider-tests` has failed on every PR since 2026-08-21. `step-basic` asserts one operation and receives `[]`, because the durable execution times out after 60s having recorded nothing.

The cause is the Managed Instances runtime, not our code, our VPC, or the SDK. Runtime `nodejs:24.DurableFunction.v33` has no outbound network connectivity from the execution sandbox: DNS resolves, but TCP/TLS to every AWS service endpoint black-holes. The SDK's first checkpoint therefore never completes — `client.send(CheckpointDurableExecutionCommand)` neither resolves nor rejects, so the `catch` that logs `CheckpointDurableExecution failed` never fires — and the execution dies at its `ExecutionTimeout`.

This PR pins capacity-provider functions to runtime `v29`, which is unaffected.

## Evidence

Same function, same capacity provider, same VPC, same SDK, us-west-2. Only the runtime differs.

Runtime `v33` (auto):

```
NET_TCP  <lambda-endpoint>:443 -> TIMEOUT
Creating checkpoint batch
(59.5s gap)
platform.report status=timeout
```

Runtime `v29` (pinned):

```
initStart runtimeVersion=nodejs:24.DurableFunction.v29
NET_TCP  18.246.196.22:443 -> ok 1ms
Creating checkpoint batch
Updating stepData from checkpoint response   (102ms round trip)
AFTER_STEP step completed
platform.report status=success
```

## What this is not

Each of these was proposed and then tested during the investigation:

- **Not VPC egress.** Reproduces with a public subnet + internet gateway (instances do get public IPs), with a private subnet + NAT gateway, and with a Lambda interface VPC endpoint — all three options in the [Managed Instances networking docs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-managed-instances-networking.html).
- **Not durable-specific.** A plain non-durable function on the same capacity provider cannot complete a TLS handshake to `lambda`, `sqs` or `sts`, and any AWS SDK call from it hangs without settling.
- **Not regional infrastructure.** On-demand functions in the same region and account reach `sts:GetCallerIdentity` in ~30ms. Only managed-instances functions are affected.
- **Not the SDK version.** Reproduced identically on `2.1.0`, `2.2.0` and `2.3.0`.
- **Not IAM.** The same execution role runs durable functions successfully off Managed Instances.

## Implementation notes

- The pin is applied **before** `PublishVersion`, because the published version snapshots the runtime management config. Pinning afterwards has no effect on the version that serves invocations.
- Scoped to `us-west-2` only. Runtime ARNs are region-scoped and only us-west-2 is verified; other regions log a message and use the default runtime.
- A failure to pin is logged but not fatal, so the deploy still succeeds and the test reports the real symptom rather than a setup error.

## Testing

- Verified live in AWS: pinning turns a 60s timeout into a ~1s success on a Managed Instances capacity provider in us-west-2. That is how the fix was found.
- `prettier --check` clean. `tsc --noEmit` reports 0 errors in `deploy-lambda.ts` and the same pre-existing error count as `main` (437, from unbuilt workspace deps).
- Not smoke-tested via the CLI locally: `npm run build -w packages/aws-durable-execution-sdk-js-examples` fails on a pre-existing `generate-examples` / `dist/examples-catalog.js` bootstrap issue that reproduces unchanged on `main`. CI exercises this path, so `capacity-provider-tests` passing on this PR is the real confirmation.

## Follow-ups (not in this PR)

- **Revert once Lambda ships a fixed runtime.** Pinning opts these functions out of automatic runtime patching and must not outlive the underlying fix. An AWS-side investigation is tracked internally.
- The capacity provider pins a single instance type (`m6g.large`), and provider logs show `InsufficientInstanceCapacity` retries in us-west-2c before succeeding in us-west-2b. Allowing more ARM types would make CI sturdier.
- Capacity is not released promptly after function deletion: `CreateFunction` can fail with `CapacityProviderScalingLimitExceeded` with zero instances running. Both providers are at `maxVCpuCount: 12` while shared across every PR run.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/scripts/deploy-lambda.ts
+++ b/packages/aws-durable-execution-sdk-js-examples/scripts/deploy-lambda.ts
@@ -22,6 +22,8 @@ import {
   State,
   LastUpdateStatusReasonCode,
   PutFunctionScalingConfigCommand,
+  PutRuntimeManagementConfigCommand,
+  UpdateRuntimeOn,
   CreateFunctionCommandInput,
 } from "@aws-sdk/client-lambda";
 import { ExamplesWithConfig } from "../src/types";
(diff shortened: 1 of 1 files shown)


yaythomas (approved):
:shipit:

Issue #865, opened 2026-08-26
NonDeterministicExecutionError is masked: invocation returns PENDING instead of FAILED


yaythomas:
## Summary

When replay validation detects non-deterministic code, the SDK terminates with `TerminationReason.CUSTOM`. `runHandler` has no case for `CUSTOM`, so the invocation falls through to the generic termination branch and returns `{Status: "PENDING"}` with no error in the response, no log line. The diagnostic message the SDK constructed ("Expected name X, but got Y") is never emitted anywhere, even with `DURABLE_VERBOSE_MODE=true`.

Returning PENDING tells the service to keep the execution alive and retry. But the purpose here is to terminate the execution: the error is deterministic and permanent, so every retry fails identically. The invocation should return `FAILED` with the error, as validation errors already do.

## Mechanism

1. [`validateReplayConsistency`](https://github.com/aws/aws-durable-execution-sdk-js/blob/6cb85ab769a7c2a356b5cc84698f375db62cf5a2/packages/aws-durable-execution-sdk-js/src/utils/replay-validation/replay-validation.ts#L22-L50) detects a Type/Name/SubType mismatch and raises `NonDeterministicExecutionError`. This is the only production error with [`terminationReason = CUSTOM`](https://github.com/aws/aws-durable-execution-sdk-js/blob/6cb85ab769a7c2a356b5cc84698f375db62cf5a2/packages/aws-durable-execution-sdk-js/src/errors/non-deterministic-error/non-deterministic-error.ts#L8).
2. [`terminateForUnrecoverableError`](https://github.com/aws/aws-durable-execution-sdk-js/blob/6cb85ab769a7c2a356b5cc84698f375db62cf5a2/packages/aws-durable-execution-sdk-js/src/utils/termination-helper/termination-helper.ts#L11-L22) terminates with that reason.
3. `runHandler` special-cases `CHECKPOINT_FAILED`, `SERDES_FAILED`, `CONTEXT_VALIDATION_ERROR`, `CONFIG_VALIDATION_ERROR` — `CUSTOM` matches none and hits the catch-all that returns [`{Status: PENDING}`](https://github.com/aws/aws-durable-execution-sdk-js/blob/6cb85ab769a7c2a356b5cc84698f375db62cf5a2/packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts#L298-L314).

[...]

PR #867, opened 2026-08-26
fix(sdk): return FAILED when replay validation detects non-determinism

ParidelPooya:
Replay validation raises NonDeterministicExecutionError, the only production error carrying TerminationReason.CUSTOM. runHandler had no branch for CUSTOM, so it fell through to the catch-all and answered {Status: "PENDING"} with no error and no log line. PENDING asks the service to keep the execution alive and invoke again, but the fault is deterministic and permanent: every replay reproduces it, the diagnostic the SDK had already built never reached the customer, and once nothing was actually pending the service rejected the response outright with "Cannot return PENDING status with no pending operations".

Invert the termination branch rather than add a third near-identical case. Suspend reasons are now an allowlist (OPERATION_TERMINATED, WAIT_SCHEDULED, RETRY_SCHEDULED, RETRY_INTERRUPTED_STEP, CALLBACK_PENDING) and every other reason answers FAILED with the carried error, so a reason added later without a branch here fails closed instead of silently claiming the execution is still progressing. This also subsumes the CONTEXT_VALIDATION_ERROR and CONFIG_VALIDATION_ERROR branches, which were copies of each other. CHECKPOINT_FAILED and SERDES_FAILED still rethrow ahead of it, unchanged.

terminateForUnrecoverableError now carries the error object, not only its message, so ErrorType reaches the customer as NonDeterministicExecutionError rather than a bare Error. This also fixes a latent bug on the CHECKPOINT_FAILED path: runHandler does `throw result.error`, which threw undefined whenever the termination came from terminateForUnrecoverableError rather than from CheckpointManager.

Tests: a composed test runs the real engine over a history checkpointed as "step-a" while the handler replays "step-b", and asserts FAILED carrying the diagnostic, with a stable-name control case. Pinned against the unfixed source it returns {Status: "PENDING"}, reproducing the report. Unit coverage added for CUSTOM -> FAILED with the type preserved, CONTEXT_VALIDATION_ERROR -> FAILED (previously untested), an unclassified reason failing closed, and all five suspend reasons -> PENDING; only OPERATION_TERMINATED was covered before.

This is the status half of #865. Item 3 of that issue -- validateReplayConsistency discarding its never-resolving promise, which lets handler code keep running against checkpoint data known to belong to a different operation -- is a separate defect with a wider blast radius, and follows in its own change.

This changes observable invocation status for a public failure mode, so it looks like a candidate for a conformance test in
https://github.com/aws/aws-durable-execution-conformance-tests.

Refs #865

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/errors/non-deterministic-error/non-deterministic-error.ts
+++ b/packages/aws-durable-execution-sdk-js/src/errors/non-deterministic-error/non-deterministic-error.ts
@@ -5,7 +5,7 @@ import { TerminationReason } from "../../termination-manager/types";
  * Error thrown when non-deterministic code is detected during replay
  */
 export class NonDeterministicExecutionError extends UnrecoverableExecutionError {
-  readonly terminationReason = TerminationReason.CUSTOM;
+  readonly terminationReason = TerminationReason.NON_DETERMINISM;
 
   constructor(message: string) {
     super(message);
--- a/packages/aws-durable-execution-sdk-js/src/termination-manager/termination-class.test.ts
+++ b/packages/aws-durable-execution-sdk-js/src/termination-manager/termination-class.test.ts
@@ -1,0 +1,52 @@
+import {
+  TERMINATION_CLASS,
+  TerminationReason,
+  classifyTermination,
+} from "./types";
+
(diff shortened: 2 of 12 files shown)


yaythomas (approved):
:shipit:

Issue #869, opened 2026-08-27
Checkpoint failure during queue drain is swallowed; execution can be reported SUCCEEDED


ParidelPooya:
## Summary

If a checkpoint batch fails while `runHandler` is draining the checkpoint queue **after** the handler promise has already won the race at L211, the failure is never observed and the invocation returns the handler's result as **SUCCEEDED**. The operation's checkpoint was discarded (`clearQueue()`), so it is neither persisted nor retried.

The root of it is that `waitForQueueCompletion()` has no way to report a failure to its caller, and the one signal that *is* raised — a termination-promise resolution — arrives after `runHandler` has already committed to the race outcome.

## Mechanism

**1. `waitForQueueCompletion()` can only resolve.** It captures `resolve` and never `reject` (`src/utils/checkpoint/checkpoint-manager.ts`):

```ts
async waitForQueueCompletion(): Promise<void> {
  if (this.queue.length === 0 && !this.isProcessing) return;
  return new Promise<void>((resolve) => { this.queueCompletionResolver = resolve; });
}
```

**2. It is resolved on the failure path too.** `processQueue`'s `catch` classifies the error, discards the queue, and raises termination; the `finally` then resolves the waiter regardless of outcome:

```ts
} catch (error) {
  const checkpointError = this.classifyCheckpointError(error);
  this.clearQueue();                               // pending checkpoints discarded
  this.terminationManager.terminate({
    reason: TerminationReason.CHECKPOINT_FAILED,
    message: checkpointError.message,
    error: checkpointError,
  });
} finally {
  this.isProcessing = false;
  if (this.queue.length > 0) { setImmediate(() => this.processQueue()); }
  else { this.notifyQueueCompletion(); ... }        // resolves waitForQueueCompletion()
}
```

**3. So the `try/catch` around the drain in `runHandler` cannot catch this** (`src/with-durable-execution.ts` L221-226). There is no rejection to catch; the `catch` is reachable only for a synchronous throw from the call itself:

```ts
try {
  await durableExecution.checkpointManager.waitForQueueCompletion();
} catch (error) {
  log("⚠️", "Error waiting for checkpoint completion:", error);
}
```

**4. The only failure signal is the termination promise, and it arrives too late.** `runHandler` commits to the outcome at L211, before the drain:

```ts
const [resultType, result] = await Promise.race([handlerPromise, terminationPromise]);
```

[...]

PR #870, opened 2026-08-27
fix(sdk): surface checkpoint failure raised after the handler resolves

ParidelPooya:
A checkpoint batch that fails while `runHandler` drains the checkpoint
queue was silently dropped when the handler had already won the race, and
the invocation returned SUCCEEDED for an operation whose checkpoint was
classified as failed and then discarded.

Neither channel out of that drain reports the failure. On the failure path
`processQueue`'s catch calls `clearQueue()`, which notifies the queue-
completion waiter, so the `try/catch` around `waitForQueueCompletion()`
sees an ordinary completion -- the waiter captures only `resolve` and is
never rejected. (The later `notifyQueueCompletion()` in `finally` is a
no-op by then, since the resolver has already been nulled.) The other
signal, `terminate({ reason: CHECKPOINT_FAILED })`, is raised after
`Promise.race` has settled: when the handler won it, nothing is awaiting
the termination promise, so the resolution has no observer, and every
branch that follows was gated on `resultType === "termination"`.

Because a result is being returned there is no subsequent invocation, so
the discarded checkpoint is never retried, contrary to the "will be
retried on next invocation" reasoning at the sibling drain sites.

Reachable from ordinary awaited code: run-in-child-context-handler is the
only handler that does not await its terminal checkpoint (SUCCEED and FAIL
are both fire-and-forget), so `await context.runInChildContext(...)` as
the last operation returns to the handler with that checkpoint still
queued. The end-to-end composed test added here is exactly that -- a plain
handler and a plain 503 -- and on main it resolves
`{"Status":"SUCCEEDED","Result":"\"finished\""}`.

Record the classified error on CheckpointManager, which already has it in
that catch, and expose `getBatchFailure()`. The caller reads the failure
from the object that produced it: no reason filter, and no dependence on
TerminationManager's `isTerminated` guard, which keeps only the first
termination raised and would hide a checkpoint failure whenever anything
else terminated first. The assignment is synchronous within that catch,
which is what matters -- a caller inside `waitForQueueCompletion()` resumes
in a microtask, so only an intervening `await` could let it observe the
failure too late.

Checked only on the handler-wins path. A suspend-class termination cannot
leave a batch to drain, since `executeTermination` gates on
`shouldTerminate()`, whose first rules require an idle queue with no
pending force requests. A fault-class termination is unguarded and can win
the race with work still queued, but already returns FAILED carrying its
own error, which is the more informative one. The residual case is a
fault-class execution-scope error masking a discarded invocation-scope
transient, which returns FAILED instead of retrying; preferring the fault
error is defensible and it is left as is.

Behaviour change beyond the drain window: on the handler-wins path any
batch failure recorded during the invocation now throws, not only one that
lands during the final drain. That matches what the termination-wins path
has always done for CHECKPOINT_FAILED.

Only partly closes this class of bug. The `serializedSize >
LAMBDA_RESPONSE_SIZE_LIMIT` branch enqueues an EXECUTION/SUCCEED
checkpoint, awaits it, drains, and returns SUCCEEDED, carrying the same
"retried on next invocation" comment. On batch failure `processQueue`'s
catch neither resolves nor rejects the in-flight batch's items and
`clearQueue()` drops queued ones silently, so the `checkpoint()` awaited
there has no settled outcome to observe. Pre-existing and left alone here; tracked as #871.

Tests: `checkpoint-batch-failure-end-to-end.composed.test.ts` drives real
withDurableExecution and a real CheckpointManager through the public API
with only the transport faked, plus a passing-drain control;
`checkpoint-batch-failure.composed.test.ts` covers the recording itself,
including that a failure is still reported when something else terminated
first; and unit tests cover the wiring in withDurableExecution.

Refs #869

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts
+++ b/packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts
@@ -72,6 +72,16 @@ export class CheckpointManager implements Checkpoint {
     reject: (error: Error) => void;
   }> = [];
   private queueCompletionResolver: (() => void) | null = null;
+  /**
+   * The classified error from a checkpoint batch that failed, if one did.
+   *
+   * Neither {@link waitForQueueCompletion}, which only ever resolves, nor the
+   * TerminationManager, whose `isTerminated` guard keeps only the first
+   * termination raised, reports this reliably.
+   */
+  private batchFailure?:
+    | CheckpointUnrecoverableInvocationError
+    | CheckpointUnrecoverableExecutionError;
   private readonly MAX_PAYLOAD_SIZE = 750 * 1024; // 750KB in bytes
   private readonly MAX_ITEMS_IN_BATCH = 250;
   private isTerminating = false;
@@ -442,6 +452,14 @@ export class CheckpointManager implements Checkpoint {
 
       const checkpointError = this.classifyCheckpointError(error);
 
+      // Assigned synchronously within this block, before clearQueue() resolves
+      // anyone inside waitForQueueCompletion(). Awaiting anything ahead of this
+      // point would let such a caller resume while the failure is still
+      // unreadable, and report success. The relative order of these two
+      // statements does not matter -- the waiter resumes in a microtask, after
+      // this block finishes -- but an await between them would.
+      this.batchFailure = checkpointError;
+
       // Clear remaining queue silently - we're terminating
       this.clearQueue();
 
@@ -467,6 +485,19 @@ export class CheckpointManager implements Checkpoint {
         }
       }
     }
+  }
+
+  /**
+   * The classified error from a checkpoint batch that failed during this
+   * invocation, or `undefined` if none did. A caller that awaited
+   * {@link waitForQueueCompletion} needs this to tell a completed drain from a
+   * failed one, since that await resolves either way.
+   */
+  getBatchFailure():
+    | CheckpointUnrecoverableInvocationError
+    | CheckpointUnrecoverableExecutionError
+    | undefined {
+    return this.batchFailure;
   }
 
   private notifyQueueCompletion(): void {
--- a/packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-batch-failure-end-to-end.composed.test.ts
+++ b/packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-batch-failure-end-to-end.composed.test.ts
@@ -1,0 +1,151 @@
+/**
+ * The handler-wins path: the customer handler returns normally while a checkpoint
+ * is still queued, and that checkpoint's batch then fails during the drain.
+ *
+ * run-in-child-context-handler does not await its terminal SUCCEED checkpoint, so
+ * `await context.runInChildContext(...)` hands the value back to the handler with
+ * that checkpoint still in the queue. The handler returns, wins the race against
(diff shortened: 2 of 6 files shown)


packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts
```diff
@@ -469,6 +487,19 @@ export class CheckpointManager implements Checkpoint {
     }
   }
 
+  /**
+   * The classified error from a checkpoint batch that failed during this
+   * invocation, or `undefined` if none did. A caller that awaited
+   * {@link waitForQueueCompletion} needs this to tell a completed drain from a
+   * failed one, since that await resolves either way.
+   */
+  getBatchFailure():
```
yaythomas, line 496: Consider adding `getBatchFailure()` to the `Checkpoint` interface (`checkpoint-helper.ts`) — as an optional method if you don't want to force it on implementers, and to `testing/mock-checkpoint.ts`. It compiles today only because `runHandler` holds the concrete `CheckpointManager`, and this PR had to hand-patch five inline `mockImplementation`s; putting it on the contract lets the type-checker keep shared mocks in sync instead.

packages/aws-durable-execution-sdk-js/src/utils/checkpoint/checkpoint-manager.ts
```diff
@@ -442,6 +452,14 @@ export class CheckpointManager implements Checkpoint {
 
       const checkpointError = this.classifyCheckpointError(error);
 
+      // Assigned synchronously within this block, before clearQueue() resolves
+      // anyone inside waitForQueueCompletion(). Awaiting anything ahead of this
+      // point would let such a caller resume while the failure is still
+      // unreadable, and report success. The relative order of these two
+      // statements does not matter -- the waiter resumes in a microtask, after
+      // this block finishes -- but an await between them would.
+      this.batchFailure = checkpointError;
```
yaythomas, line 461: Nit: last-write-wins if more than one batch fails in the same invocation. On the handler-wins path nothing sets `isTerminating`, so after `clearQueue()` a later checkpoint can still enqueue, fail, and overwrite the recorded error with a less relevant one. `this.batchFailure ??= checkpointError;` would keep the first failure, which is the likelier root cause.

yaythomas (commented):
Nice!

Could file a follow-up issue for the acknowledged residual (the `serializedSize > LAMBDA_RESPONSE_SIZE_LIMIT` path, where the awaited `checkpoint()` never settles on batch failure and the queued items are dropped silently), so it doesn't get lost? the PR description documents it well but a description is not trackable.

ParidelPooya:
Filed as #871.

Turned out slightly broader than the description had it, so worth restating: the batch is spliced out of the queue *before* being sent (`checkpoint-manager.ts:408`), so on failure the catch never touches `batch` at all — those `checkpoint()` promises are neither resolved nor rejected, not just the ones `clearQueue()` drops.

That is survivable everywhere except the large-result path, because every other awaited checkpoint is reachable only from handler code, where the termination promise wins the race and abandons the handler. The large-result path runs after the race has settled, so there is nothing left to abandon it and the await has no timeout — it stalls until the platform kills the invocation rather than failing fast.

The issue notes the two fix directions and why rejecting the batch is not a one-liner: `checkpoint()` already returns a deliberately never-resolving promise when terminating or when an ancestor has finished, so unconditional rejection would surface unhandled rejections at call sites built around being abandoned.

Issue #872, opened 2026-08-28
Flaky: ADOT execution-view X-Ray E2E misses durable operation spans (fails on main too)


ParidelPooya:
### Summary

The ADOT execution-view X-Ray end-to-end check fails intermittently because the durable operation spans are absent from the trace returned by `GetTraceSummaries`/`BatchGetTraces`. The workflow itself runs and succeeds — only the span assertions fail. It reproduces on `main`, not just on pull requests.

Two CI checks fail together from this one cause:

- `integration-tests (22.x) / Jest Integration Tests`, via `packages/aws-durable-execution-sdk-js-examples/src/examples/otel/adot-execution-xray-e2e/otel-adot-execution-xray-e2e.test.ts`
- `opentelemetry / Execution view / Execution - ADOT/X-Ray`, orchestrated from the conformance-tests repository

### Observed failure

```
● otel-adot-execution-xray-e2e (cloud) › should execute workflow and produce traces via ExecutionOtelPlugin + ADOT layer

  Missing span names in X-Ray trace: [fetch-data, short-pause, process-data,
  child-operations, inner-step, fails-then-succeeds].
  Found spans: [<fn>, Invocation, <fn>, <fn>, Invocation, <fn>, <fn>, Overhead,
  <fn>, Overhead, <fn>, <fn>, <fn>, <fn>, Overhead, <fn>, Overhead, Init, <fn>,
  Invocation, <fn>, Invocation]

  at src/examples/otel/adot-execution-xray-e2e/otel-adot-execution-xray-e2e.test.ts:43

● Test suite failed to run
  assertEventSignature was not called for test otel-adot-execution-xray-e2e
```

Every span present is Lambda-generated (`Init`, `Invocation`, `Overhead`, and the function segment). None of the plugin-generated operation subsegments are there. The second error is a cascade: `assertSpanNames` throws before the test reaches `assertEventSignatures`, and the teardown check then reports it as never called.

The assertions preceding it pass, so the execution reached `SUCCEEDED`, returned the expected step results, and produced a well-formed `X-Amzn-Trace-Id` header with a valid 32-hex trace ID. Only the subsegments are missing.

### Intermittent, and present on main

Recent `OTel Conformance Tests` conclusions on `main`, newest first:

```
failure  failure  failure  failure  success  success  failure  failure  success  failure
```

[...]

PR #873, opened 2026-08-28
test(examples): park flaky ADOT X-Ray span assertions

ParidelPooya:
The ADOT execution-view end-to-end test fails intermittently because the
durable operation subsegments are absent from the trace returned for the
execution: only the Lambda-generated segments (Init, Invocation, Overhead
and the function itself) come back, so every expected span name is
reported missing.

It is not caused by any change under test. The OTel conformance workflow
shows the same failure on main, including on the current tip, interleaved
with passes; and 24.x passes these assertions in the same runs where 22.x
fails, which points away from a code-level cause and towards the exported
subsegments not being queryable yet when the trace is fetched.

Guard only the span-shape assertions, so the test keeps checking that the
workflow reaches SUCCEEDED, that the step results are correct, and that a
well-formed X-Amzn-Trace-Id header with a valid trace ID is produced. The
preceding assertions were already passing, so this parks the coverage that
cannot currently pass rather than the whole test.

A single named constant rather than `it.skip` keeps the assertions visible
and greppable, and makes re-enabling a one-line change.

The sibling `Execution - ADOT/X-Ray` check is orchestrated from the
conformance-tests repository and is unaffected by this; it fails from the
same cause and is covered by the same issue.

Refs #872

### DIFF
--- a/packages/aws-durable-execution-sdk-js-examples/src/examples/otel/adot-execution-xray-e2e/otel-adot-execution-xray-e2e.test.ts
+++ b/packages/aws-durable-execution-sdk-js-examples/src/examples/otel/adot-execution-xray-e2e/otel-adot-execution-xray-e2e.test.ts
@@ -10,6 +10,14 @@ import {
   extractTraceIdFromXRayHeader,
 } from "../../../utils/xray-trace-helper";
 
+/**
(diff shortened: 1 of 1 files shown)


ayushiahjolia:
No longer needed after https://github.com/aws/aws-durable-execution-sdk-js/pull/876

PR #874, opened 2026-08-31
fix(sdk): preserve cross-realm errors with a realm-safe isErrorLike check

### LINKED ISSUE #306: [Bug] Errors from different realms are not preserved
Since we use `instanceof Error` checks in our code, if an Error does not come from the same realm (e.g. a separate vm context or a worker), then it will fail the check and pass a generic error.

We should probably use an `isErrorLike` function or something instead of checking `instanceof`, and also sweep the codebase for other uses of `instanceof` to make sure it will work fine with different realms.

Example:

```typescript
import {
  DurableContext,
  withDurableExecution,
} from "@aws/durable-execution-sdk-js";
import { ExampleConfig } from "../../../types";
import * as vm from "vm";

export const config: ExampleConfig = {
  name: "Cross Realm Handler Error",
  description:
    "Demonstrates how errors thrown from different Node.js execution realms are captured and structured in results",
};

export const handler = withDurableExecution(
  async (event: any, context: DurableContext) => {
    await context.step(
      async () => {
        // Create a new V8 context (different realm) and throw an error from it
        const vmContext = vm.createContext();

        // Execute code in the different realm that throws an error
        const code = `
      throw new Error("Error from different Node.js realm");
    `;

        // This will throw an error from a different V8 context/realm
        vm.runInContext(code, vmContext);
      },
      {
        retryStrategy: (error, attemptCount) => {
          if (error.message.includes("realm")) {
            return {
              shouldRetry: false,
            };
          }
          return {
            shouldRetry: attemptCount < 3,
          };
        },
      },
    );
  },
);
```

```typescript
import { handler } from "./cross-realm-error";
import { createTests } from "../../../utils/test-helper";
import { ExecutionStatus } from "@aws/durable-execution-sdk-js-testing";

createTests({
  name: "cross-realm-error test",
  functionName: "cross-realm-error",
  handler,
  tests: (runner) => {
    it("should handle cross-realm errors gracefully and capture error details", async () => {
      const result = await runner.run({
        payload: { test: "cross-realm-error-case" },
      });

      // Check that error was captured in the result
      const error = result.getError();
      expect(error).toEqual({
        errorMessage: "Error from different Node.js realm",
        errorType: "StepError",
        stackTrace: undefined,
      });

      expect(result.getOperations()).toHaveLength(1);
      expect(runner.getOperationByIndex(0).getStepDetails()?.attempt).toBe(1);

      expect(result.getStatus()).toBe(ExecutionStatus.FAILED);
    });
  },
});

```

manishh-13:
## Problem
Closes #306. Error detection and normalization used `error instanceof Error`, which only matches the current realm's `Error` constructor. Errors thrown from a different realm (a `vm` context, a `worker_threads` worker, etc.) fail that check and were flattened to `new Error("Unknown Error")` / `new Error(String(error))` before reaching the user's retryStrategy, plugin hooks, and serialization, discarding the original message, name, and stack.

## Fix
- Extract the existing duck-typed `isErrorLike` (previously private in `error-object.ts`) into a shared `utils/error-object/is-error-like.ts` helper. Semantics are unchanged, so `createErrorObjectFromError` behaves identically; the helper is now reusable and documented.
- Replace raw `instanceof Error` with `isErrorLike` at every site that captures or serializes a customer-originated error: step handler (retryStrategy + plugin hooks), wait-for-condition handler, concurrent-execution handler, promise handler (`allSettled`), serdes errors, checkpoint manager, plugin loader, default logger, safe-stringify, the `DurableOperationError` cause walk, and batch-result serialization.
- Left unchanged by design: SDK-internal class checks (`DurableOperationError`, `ChildContextError`, the unrecoverable-error guards, the `isDurableExecutionClientError` public type guard) and the user-supplied class match in `retry-common` (`error instanceof ErrorType`), which are always same-realm.

## Testing
- New unit tests for `isErrorLike`, including an actual cross-realm error thrown from a `vm` context.
- New regression test in `step-handler.test.ts`: a step that throws a cross-realm error now forwards the original error (with its real message) to `retryStrategy` instead of `new Error("Unknown Error")`. Verified as a true regression: it fails on the pre-fix code and passes with the fix.
- Full SDK package suite passes (105 suites / 1253 tests), ESLint clean. No conformance tests added, per CONTRIBUTING.

Happy to add an examples-package example (as sketched in the issue) if you'd prefer end-to-end/local-runner coverage.

### DIFF
--- a/packages/aws-durable-execution-sdk-js/src/utils/logger/default-logger.ts
+++ b/packages/aws-durable-execution-sdk-js/src/utils/logger/default-logger.ts
@@ -5,6 +5,7 @@ import {
   DurableLogger,
   DurableLoggingContext,
 } from "../../types/durable-logger";
+import { isErrorLike } from "../error-object/is-error-like";
 
 export interface LoggingExecutionContext {
   durableExecutionArn: string;
@@ -83,7 +84,7 @@ function jsonErrorReplacer(
   _key: string,
   value: DurableLogField,
 ): DurableLogField {
-  if (value instanceof Error) {
+  if (isErrorLike(value)) {
     return Object.assign(
       {
         errorType: value?.constructor?.name ?? "UnknownError",
@@ -173,7 +174,7 @@ function formatDurableLogData(
 
   result.message = formatWithOptions(FORMAT_OPTIONS, ...messageParams);
   for (const param of messageParams) {
-    if (param instanceof Error) {
+    if (isErrorLike(param)) {
       result.errorType = param?.constructor?.name ?? "UnknownError";
       result.errorMessage = param.message;
       result.stackTrace =
--- a/packages/aws-durable-execution-sdk-js/src/utils/safe-stringify/safe-stringify.ts
+++ b/packages/aws-durable-execution-sdk-js/src/utils/safe-stringify/safe-stringify.ts
@@ -1,3 +1,4 @@
+import { isErrorLike } from "../error-object/is-error-like";
 export const safeStringify = (data: unknown): string => {
   try {
     const seen = new WeakSet();
@@ -9,7 +10,7 @@ export const safeStringify = (data: unknown): string => {
           seen.add(value);
 
           // Handle Error objects by extracting their properties
-          if (value instanceof Error) {
+          if (isErrorLike(value)) {
             return {
               ...value,
               name: value.name,
--- a/packages/aws-durable-execution-sdk-js/src/errors/durable-error/durable-error.ts
+++ b/packages/aws-durable-execution-sdk-js/src/errors/durable-error/durable-error.ts
@@ -2,6 +2,7 @@ import { ErrorObject } from "../../types/wire";
 import { STORE_STACK_TRACES } from "../../utils/constants/constants";
 // Type-only import to avoid a runtime circular dependency with types/batch.
 import type { CompletionReason } from "../../types/batch";
+import { isErrorLike } from "../../utils/error-object/is-error-like";
 
 /**
  * Base class for all durable operation errors
@@ -115,7 +116,7 @@ export abstract class DurableOperationError extends Error {
           errorData = node.errorData;
           break;
         }
-        node = node instanceof Error ? node.cause : undefined;
+        node = isErrorLike(node) ? node.cause : undefined;
       }
     }
     return {
--- a/packages/aws-durable-execution-sdk-js/src/errors/serdes-errors/serdes-errors.ts
+++ b/packages/aws-durable-execution-sdk-js/src/errors/serdes-errors/serdes-errors.ts
@@ -3,6 +3,7 @@ import { TerminationManager } from "../../termination-manager/termination-manage
 import { UnrecoverableInvocationError } from "../unrecoverable-error/unrecoverable-error";
 import { log } from "../../utils/logger/logger";
 import { SerdesContext } from "../../utils/serdes/serdes";
+import { isErrorLike } from "../../utils/error-object/is-error-like";
 
 /**
  * Error thrown when serdes operation fails and terminates Lambda invocation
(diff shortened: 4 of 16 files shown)


packages/aws-durable-execution-sdk-js/src/utils/logger/default-logger.ts
```diff
@@ -173,7 +174,7 @@ function formatDurableLogData(
 
   result.message = formatWithOptions(FORMAT_OPTIONS, ...messageParams);
   for (const param of messageParams) {
-    if (param instanceof Error) {
+    if (isErrorLike(param)) {
```
hln33: Would the original code be better in this instance?

---

This can cause some logs to have error fields incorrectly injected into them.

For instance, if a user uses the logger like:
```
logger.log(
    DurableLogLevel.INFO, 
    "customer data", 
    { name: "bob", message: "please gift wrap" }
);
```
`isErrorLike()` would evaluate true and add fields like `errorType` and `errorMessage` to `result`. This can cause confusion for those inspecting logs and seems like a regression.

packages/aws-durable-execution-sdk-js/src/utils/logger/default-logger.ts
```diff
@@ -83,7 +84,7 @@ function jsonErrorReplacer(
   _key: string,
   value: DurableLogField,
 ): DurableLogField {
-  if (value instanceof Error) {
+  if (isErrorLike(value)) {
```
hln33: Similar to comment below. This can cause some logs to have error fields incorrectly injected into them.

manishh-13:
Good catch, thank you. You are right, and it was a regression I introduced. Your example is exactly the failure:

```js
logger.log(DurableLogLevel.INFO, "customer data", { name: "bob", message: "please gift wrap" });
```

Looking at it properly, the root cause was my refactor rather than the predicate itself. That permissive `message`/`name` duck-type already existed on `main`, but it was **private to `error-object.ts`**, where the leniency is deliberate and covered by tests (`createErrorObjectFromError` accepts an error-shaped payload on its way to the wire). What I did wrong was hoist that one function into a shared util and reuse it at every site that previously said `instanceof Error`, including the logger, where the leniency becomes a bug.

So rather than reverting the two logger sites, I split it into two predicates that say what they mean:

**1. `isError` (new shared util, strict and realm-safe)** replaces the old `instanceof Error` checks:

```ts
export function isError(value: unknown): value is Error {
  return (
    value instanceof Error ||
    Object.prototype.toString.call(value) === "[object Error]"
  );
}
```

The `[object Error]` brand comes from the `ErrorData` internal slot that only the `Error` constructor can create, and that slot is realm independent. So it still catches errors from a `vm` context or a worker, which was the point of the PR, while `{ name: "bob", message: "please gift wrap" }` is now correctly `false`. A value can opt in deliberately with `[Symbol.toStringTag] = "Error"`, which ordinary data does not do by accident.

**2. `error-object.ts` keeps its own local `isErrorLike`**, restored to the `main` behaviour and now layered on `isError`. Its doc comment states why the leniency is intentional there and why it must not leak into logging. That keeps the pre-existing, tested wire-serialization behaviour untouched.

Results for the cases that matter:

| value | before this push | now |
|---|---|---|
| same-realm `Error` and subclasses | true | true |
| `Object.create(Error.prototype)` | true | true |
| cross-realm `Error` and `TypeError` | true | true |
| `{ name: "bob", message: "please gift wrap" }` in the **logger** | true (your bug) | false |
| `{ name, message }` in **wire serialization** | true | true (unchanged from `main`) |

I added your example as a named regression test, plus cases for a stack-carrying object, a null-prototype object, cross-realm subclasses and the `Symbol.toStringTag` opt-in. The two `error-object` tests that started failing when I first tightened the shared predicate are the ones that caught this design problem, and they now pass unmodified, which I think is the right signal: the pre-existing contract is preserved and only the logging paths got stricter.

Let me know if you would rather I just revert the two logger sites and keep the change narrower; happy to go that way.

packages/aws-durable-execution-sdk-js/src/utils/safe-stringify/safe-stringify.ts
```diff
@@ -9,7 +10,7 @@ export const safeStringify = (data: unknown): string => {
           seen.add(value);
 
           // Handle Error objects by extracting their properties
-          if (value instanceof Error) {
+          if (isErrorLike(value)) {
```
hln33: Similar to comments in `default-logger.ts`, this method takes in user-passed-in data and is therefore at risk of wrongly classifying objects as errors.

Would prefer the original code here.

hln33 (changes requested):
Most of the changes to use `isErrorLike()` look good to me. Just have some concerns regarding its usage inside logger files and safe-serialize.

Issue #854, opened 2026-08-21
otel: join Workflow to propagated trace and restore operation links


zhongkechen:
## Summary

Implement the revised shared OpenTelemetry topology tracked by
aws/aws-durable-execution-conformance-tests#99.

`_X_AMZN_TRACE_ID.Root` defines the canonical execution trace when valid.
However, `_X_AMZN_TRACE_ID.Parent` is only the immediate remote context exposed
to the Lambda runtime and is not guaranteed to be the durable backend attempt
span directly.

Create `Workflow` in the canonical trace and allow either of these parents:

- the remote context reconstructed from `_X_AMZN_TRACE_ID.Parent`; or
- the current valid ambient span when it has the same trace ID.

ADOT and community Node.js auto-instrumentation layers can create the ambient
handler span before the durable plugin runs. OpenTelemetry `SpanContext` has no
ancestor pointer, so the plugin cannot traverse from that local span to the
durable backend span. Same-trace membership is the available portable check.

## Required changes

- Parse `Root`, `Parent`, and `Sampled` independently; only `Sampled=0` and
  `Sampled=1` are authoritative upstream decisions.
- Create `Workflow` with its deterministic execution-scoped span ID in the
  canonical trace.
- Allow `Workflow` to be parented to either the reconstructed remote context or
  the current valid same-trace ambient context.
- Do not require or claim that the reconstructed `Parent` is directly the span
  named `Durable Execution Attempt #1`.
- For `Invocation`, prefer the active ambient span only when its trace ID
  matches the canonical trace; otherwise use the reconstructed remote parent.
- Do not attempt to traverse an ambient span's parent chain; the OTel
  JavaScript API does not expose one.
- When no valid remote parent can be constructed, ignore unrelated ambient
  context and parent `Workflow` and all `Invocation` spans to the deterministic
  synthetic execution root.
- Restore continuation and replay links using the canonical trace ID and
  `deriveSpanIdFromOperationId(operationId, executionArn)`.
- Retain the Workflow and Invocation correlation links required by the shared
  contract.
- Stop hard-coding extracted or reconstructed contexts as sampled.

## Valid parent shapes

```text
Propagated remote parent
├── Workflow
└── Same-trace ambient Lambda span
    └── Invocation

Propagated remote parent
└── Same-trace ambient Lambda span
    ├── Workflow
    └── Invocation
```

The selected Workflow parent can be absent from a backend query. If it is
[...]

PR #880, opened 2026-09-01
feat(otel): wire both plugins onto the shared execution trace

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-js/issues/854

*Description of changes:*
- Third PR in the stack, and the behavior flip. 
- Wires both InvocationOtelPlugin and ExecutionOtelPlugin onto the shared execution trace and sampling from PR 2. 
- The Workflow and Invocation spans now share one trace, anchored to a propagated backend parent when present or a deterministic synthetic execution root otherwise; an ambient Lambda span is used as the Invocation parent only when it is already on the execution trace and carries the same sampled bit. 
- Sampling is resolved once per invocation and applied to every durable span via DurableSampler. Operation and attempt spans link to the Workflow span for execution-level correlation, and cross-invocation replay/continuation segments link back to the deterministic initial operation span. 
- Next in stack: exports and docs.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-otel/src/invocation-plugin.ts
+++ b/packages/aws-durable-execution-sdk-js-otel/src/invocation-plugin.ts
@@ -21,22 +21,28 @@ import {
   trace,
   SpanKind,
   SpanStatusCode,
-  TraceFlags,
   ROOT_CONTEXT,
   isSpanContextValid,
+  TraceFlags,
 } from "@opentelemetry/api";
 import { hrTime } from "@opentelemetry/core";
+import { SamplingDecision } from "@opentelemetry/sdk-trace-node";
 import {
   DeterministicIdGenerator,
-  deriveTraceIdFromArn,
   deriveWorkflowSpanId,
   deriveSpanIdFromOperationId,
 } from "./deterministic-id-generator";
 import { xRayContextExtractor } from "./context-extractors";
 import type { ContextExtractor } from "./context-extractors";
+import {
+  canonicalTraceId,
+  resolveExecutionTraceContext,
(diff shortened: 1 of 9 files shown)


packages/aws-durable-execution-sdk-js-otel/src/invocation-plugin.ts
```diff
@@ -314,9 +330,76 @@ export class InvocationOtelPlugin implements DurableInstrumentationPlugin {
...
+   * Because the initial operation span's ID is reproducible, this restores the
+   * cross-invocation link that was previously dropped as "fabricated" — it now
+   * targets a genuinely exported span (open operation spans are ended on
+   * non-terminal invocations; see onInvocationEnd).
+   */
+  private initialOperationLink(operationId: string): Link | undefined {
+    if (!this.executionTraceId) {
+      return undefined;
+    }
+    const initialContext: SpanContext = {
```
zhongkechen, line 394: Does this work for virtual context?

zhongkechen (dismissed):

packages/aws-durable-execution-sdk-js-otel/src/invocation-plugin.ts
ayushiahjolia: Yes, virtual context works. Parenting comes from info.parentId, so child-context/parallel/map ops form a correct subtree under the Workflow span regardless of virtual vs real boundaries. The once-per-execution sampling decision propagates into the subtree through DurableSampler's AsyncLocalStorage.

zhongkechen (approved):

zhongkechen (approved):

PR #882, opened 2026-09-01

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-js/issues/831

*Description of changes:*
- Workflow span: created+ended once, on the terminal invocation (both plugins).
- Attempt span: created+ended in the same invocation (both plugins).
- Operation span: invocation plugin → one segment per invocation (fresh ID + link across invocations); execution plugin → one span total, created+ended in the invocation where the operation terminates (held as a non-recording placeholder while it spans invocations).

### DIFF
--- a/packages/aws-durable-execution-sdk-js-otel/README.md
+++ b/packages/aws-durable-execution-sdk-js-otel/README.md
@@ -205,12 +205,13 @@ Attempt   -> Invocation
 
 The plugin makes Workflow the active span while durable handler code runs.
 Completed operation spans use durable operation start/end timestamps and
-deterministic span IDs. An operation that remains open when an invocation
-suspends is not ended or exported; a later invocation recreates and exports it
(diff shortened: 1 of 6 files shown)


zhongkechen:
Updated with 2652898f.

Root cause: ExecutionOtelPlugin deferred recording operation spans until onOperationEnd, but normal operation-start hooks can omit the backend start timestamp. That caused the parent operation span to default to the end-hook time and begin after its attempt or child spans.

Fix: capture the operation-start hook time when the backend timestamp is absent, preserve the earliest observation across repeated starts, and add regression coverage that asserts the deferred operation contains its attempt.

Validation:
- OTel package tests: 336/336 passed
- OTel package lint: passed
- OTel package build: passed
- git diff --check: passed

wangyb-A (approved):

wangyb-A (approved):

PR #883, opened 2026-09-01
build: replace ESLint and Prettier with Biome

ParidelPooya:
Replaces ESLint and Prettier with [Biome](https://biomejs.dev) for JavaScript, TypeScript and JSON. Prettier is dropped outright rather than retained for Markdown and YAML — see below. The published `@aws/durable-execution-sdk-js-eslint-plugin` is unaffected and still built, tested and run against the examples — customers keep using it with ESLint.

One toolchain instead of two, one config instead of six, and lint runs in CI for the first time as a dedicated job.

## What changes

**Removed:** 4 ESLint configs (the examples one is rewritten to run only the published plugin), `.prettierrc`, `.prettierignore`, the `prettier` devDependency, the local `eslint-plugin-filename-convention` package, and the `lint` step from three `prebuild` hooks. ESLint remains in exactly two places: the plugin package (its `RuleTester` tests need it) and the examples package (which runs the plugin's own `no-nested-durable-operations` rule).

**Added:** `biome.jsonc` — JSONC so the rationale from the deleted configs survives as comments rather than being lost. `.github/workflows/lint.yml` with a single `lint` job, wired into `build.yml`.

**Scripts:**

| Script | Purpose |
| --- | --- |
| `lint` / `lint:fix` | Biome check, with `--diagnostic-level=error` on `lint` so a failure names its file |
| `lint:ci` | `biome ci`, the gate |
| `lint:suppressions` | Fails on any suppression that suppresses nothing |
| `lint:versions` | Fails if the seven exact `@biomejs/biome` pins disagree |
| `lint:filenames` | Kebab-case **directory** names under the SDK's `src`, which Biome cannot check |
| `lint:async-safety` | Annotates the three async rules on changed files only |
| `lint:report` | Renders the warning backlog into the job summary; never gates |

**Speed:** lint went from ~0.5s to ~3s, entirely from enabling Biome's `types` domain to recover the async-safety rules. Worth it.

## Enforcement: what was gated before, and what is gated now

ESLint was **not** a pre-commit-only concern. The core SDK and testing packages ran lint in `prebuild` — the testing package through both `strictTypeChecked` and eslint-plugin-jest's recommended set — and CI runs `npm run build`, so their error-level rules failed every PR. Full accounting:

| Before | After |
| --- | --- |
| `filename-convention/kebab-case` — error, gated, checked filenames **and directory segments** | `useFilenamingConvention` for filenames + `lint:filenames` for directories — both gated |
| wire-model `no-restricted-imports` — error, gated | `noRestrictedImports` — gated |
| insight host-boundary rules — error, gated via `insight-hosts` | gated via the new `lint` job |
| 4 of `typescript-eslint/recommended` — error, gated | equivalent Biome rules, already error — gate |
| 7 of `typescript-eslint/recommended` — error, gated | equivalent Biome rules existed but were enabled **below error**, so they did not gate — **now `error`** |
| 21 more from `eslint:recommended`, `strictTypeChecked` and `stylisticTypeChecked` — error, gated | equivalent Biome rules were **below error or off**, so they did not gate — **now `error`**, repo-wide where the tree is clean and in the testing package's override otherwise |
| `no-namespace`, `prefer-as-const`, `no-require-imports`, `no-unused-expressions` — error, gated | **now enabled at error, repo-wide** — broader scope than before |
| testing pkg `noFloatingPromises`, `noMisusedPromises`, `useAwaitThenable`, `only-throw-error` — error via `strictTypeChecked` | `noFloatingPromises`, `noMisusedPromises`, `useAwaitThenable`, `useThrowOnlyError` — **error, gated** via the testing-package override (all zero-violation, restored at no cost) |
| testing pkg `no-non-null-assertion` over **non-test source** — error via `strictTypeChecked` | `noNonNullAssertion` — **error, gated** over the testing package's non-test source; the four reviewed sites carry suppressions. Off repo-wide (~666 sites) |
| testing pkg `jest/no-disabled-tests` — error via eslint-plugin-jest recommended | `noSkippedTests` — **error, gated** via the testing-package override; the three skipped tests carry suppressions |
| testing pkg `require-await` — error via `strictTypeChecked` | `useAwait` exists but is **left off** — broader than ESLint's rule (19 sites vs 7), so it is on the documented triage list |
| **`no-duplicate-imports`** — error, gated | **no Biome equivalent — lost** |
| **`triple-slash-reference`** — error, gated | **no Biome equivalent — lost** |
| **rest of `strictTypeChecked`** (the `any`-propagation family, `restrict-template-expressions`, `no-confusing-void-expression`, `prefer-promise-reject-errors`, `no-unnecessary-type-parameters`) — error, gated | **no Biome equivalent — lost**; each reviewed site carries an inline note |

Net: enforcement is **broader** — every package is linted in CI now, four rules gained repo-wide scope, `noUnusedImports` is promoted to error, and the testing package's `strictTypeChecked`/jest gates that have Biome equivalents are restored (the async trio, `useThrowOnlyError`, `noNonNullAssertion` over non-test source, and `noSkippedTests`) — but **not strictly stronger**. The rules with no Biome equivalent regress, and all are recorded in `biome.jsonc` with the measurements behind them rather than left to be rediscovered; each reviewed suppression site for those rules carries an inline note. `useAwait` (the `require-await` equivalent) is broader than ESLint's rule, so it is documented on the triage list rather than enabled.

`noUnsafeTypeAssertion` would close part of the largest gap, and in the scope ESLint enforced it is near parity: of its 265 diagnostics in the testing package only 27 are in non-test source (13 files) against ESLint's 23 reviewed suppressions there; the other 238 are in `*.test.ts` files that the old config exempted (`no-unsafe-type-assertion: "off"` for tests). Left off deliberately: restoring it to the non-test scope ESLint enforced is small, but enabling it repo-wide including tests is a triage project. The reviewed set is recoverable with one `git grep` against the pre-migration commit, and the config says so.

## The warning backlog

**996 warnings and 19 infos, 0 errors.** These are rules Biome recommends that were never enforced here. Turning them on would rewrite thousands of call sites inside a toolchain migration and bury any real behavioural difference, so they are `warn` and documented per rule in `biome.jsonc` with a site count and whether a safe autofix exists. None is rejected on principle.

CI gates on errors only (`--diagnostic-level=error`). Formatting violations and lint errors are both error-severity, so both still fail; the backlog is reported in the job summary instead of annotating 996 lines the PR never touched.

**34 suppressions** (30 inline, 4 file-level). Six name a genuine defect and say why the mechanical fix is wrong; the rest carry a greppable migration reason. `lint:suppressions` fails CI on any that stops suppressing something, which the `biome ci` gate cannot do on its own: `suppressions/unused` is warning-severity, so `--diagnostic-level=error` filters it out.

## Behaviour

No runtime behaviour changes in shipped code. Two candidates were identified and reverted:

- `durable-context.ts` — dropping `return` from the logger wrappers would have stopped forwarding the concrete logger's return value, since `DurableContextLogger` is `Pick<Logger, ...>` and only the `DurableLogger` constraint declares `void`. The non-comment diff for that file against `main` is now empty.
- Em-dash escaping in Markdown, from a script that defaulted to `ensure_ascii=True`.

One autofix in that category was kept rather than reverted, and reached shipped source: the `function(){}` → arrow-function rewrite in the testing SDK's `test-runner/local/invoke-handler.ts`, which converts four Lambda-context mock methods (`getRemainingTimeInMillis`, `done`, `fail`, `succeed`). Arrow conversion is a `this`-binding change, and this is a published package — but it is safe here: none of the four methods references `this`, and `...defaultContextValues` still spreads last so any constructor-provided override still wins.

The insight `.tsx` files appear reformatted because nothing had ever formatted them — the old lint-staged glob was `*.{js,mjs,ts,md,json,yml}`, with no `tsx`. That is net-new coverage from Biome, and it explains `SettingsModal.tsx` at 377 changed lines (pure reflow, verified).

## No Markdown or YAML formatter

Biome has no formatter for either language, and Prettier was dropped rather than kept just for them. Almost nothing here hand-writes those files — the conformance CloudFormation templates, the issue templates and most docs are generated or tool-authored — so a formatter buys consistency nobody was going to violate.

Keeping Prettier for `*.{md,yml,yaml}` was tried first and rejected for three reasons:

- **`.yaml` coverage would have been new, not preserved.** The old glob had `yml`, not `yaml`, so nothing had ever formatted the 10 conformance templates. Widening it reformatted them for ~950 lines of quote-style and indentation churn — the largest single change in the PR, and entirely cosmetic.
- **It would have been a new hard gate.** There was no `prettier --check` anywhere in CI before, and no lint job at all; Prettier ran only through lint-staged, which `--no-verify` bypasses. A CI check turns a bypassable convention into a merge blocker for docs.
- **It made the Prettier version load-bearing.** `--check` passes on 3.7.0 and 3.9.6 but **fails on 3.6.2**, because Prettier changed Markdown table width computation in that range. It had already produced one near-miss: Prettier trimmed only the last of three trailing-space lines in `bug_report.yml`, which would have failed CI for anyone whose editor strips trailing whitespace on save.

Every formatting-only Markdown and YAML change is reverted, so the churn does not land. Markdown/YAML in this PR goes from 702+/540− across 21 files to 210+/52− across 8, and each remaining one is a substantive edit. Adding a formatter for these later should be its own change, with its own decision about gating.

## At merge

1. **Add `lint / lint` to branch protection.** Recommended for visibility, but no longer load-bearing: `build` declares `needs: [lint-commits, lint]`, so a lint failure blocks `build` and everything downstream of it, which is the same topology as before this branch (lint used to run inside `prebuild`). Adding the check makes the failure legible on the PR rather than showing up as a skipped build.
2. **Re-run the two red checks first.** `integration-tests (22.x)` fails on a parallel-invoke staggered-completion count that passes on 24.x, and one otel Dash0 job fails. Neither touches lint, but confirm by re-run rather than assumption.
3. Contributors with branches predating this will see formatting differences on first push. `npm run lint:fix` resolves them in one pass; `CONTRIBUTING.md` says so.

One caveat on fork PRs: `build.yml` triggers on both `pull_request` and `pull_request_target: types: [labeled, unlabeled]`. The `pull_request` runs check out the merge ref and do lint the PR, so a green check is real. Only label-triggered runs check out base. Pre-existing and shared by every job in the repo.

## Verification

`npm run build`, `npm test`, `lint:ci`, `lint`, `lint:versions`, `lint:suppressions`, `lint:report`, `lint:async-safety`, `lint:durable-rules`, the webview-ui `tsc --noEmit`, and `actionlint` on the new workflow all exit 0.

Each restored gate was verified by planting a violation and confirming `biome ci` fails, not by reading the config.


.github/workflows/build.yml
```diff
@@ -110,6 +108,12 @@ jobs:
       - name: Bundle the MCP server
         run: npm run build -w packages/aws-durable-execution-sdk-js-insight-mcp
 
+  # Runs off lint-commits rather than build: Biome needs only the sources, so
+  # linting does not have to wait for compilation and reports failures earlier.
+  lint:
```
yaythomas, line 120: The required build paths no longer run lint, while the current ruleset does not require the new standalone check. Until that is updated, a PR could merge despite Biome failures.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,163 @@
...
+        #
+        # `steps.install.outcome == 'success'` is the part that actually prevents the
+        # npm ci problem. Without it a failed install cascades into five steps that
+        # run with no node_modules and fail on a missing binary, burying the one real
+        # error. Note the implicit success() that GitHub adds to a bare `if` is
+        # suppressed once a status function appears, which is why both halves are
+        # needed. ${{ }} is required because a leading `!` is YAML tag syntax.
+        if: ${{ !cancelled() && steps.install.outcome == 'success' }}
+        run: |
+          BASE="origin/${GITHUB_BASE_REF:-main}"
```
yaythomas: On push events, GITHUB_BASE_REF is empty, so this falls back to origin/main. For pushes to v1.x, that compares against the wrong branch and may report unrelated differences.. either run this step only for pull requests or use origin/${GITHUB_REF_NAME} as the base for push events?

biome.jsonc
```diff
@@ -0,0 +1,891 @@
...
+      // cross-SDK contract with aws-durable-execution-conformance-tests, and the insight
+      // packages use camelCase throughout. Neither should be renamed by a lint rule.
+      "includes": [
+        "packages/aws-durable-execution-sdk-js/src/**/*.ts",
+        "!packages/aws-durable-execution-sdk-js/src/**/__mocks__/**"
+      ],
+      "linter": {
+        "rules": {
+          "style": {
+            "useFilenamingConvention": {
```
yaythomas, line 683: The removed custom rule checked every directory segment, whereas Biome’s useFilenamingConvention appears to validate filenames only. This means future non-kebab-case directories under the SDK’s src tree would pass lint. intentional?

.github/workflows/build.yml
ParidelPooya: Correct, and it is worse than an unenforced check — it is a change in topology. Before this branch, `npm run build` invoked eslint via `prebuild` for three packages, so a lint error failed `build` and everything downstream of it. Making `lint` a sibling job removed that.

Fixed in b09db048 by giving `build` `needs: [lint-commits, lint]`, which restores the old dependency without reviving the hooks. Adding `lint / lint` to the ruleset is still worth doing for visibility, but it is no longer the only thing preventing a merge over a Biome failure.

.github/workflows/lint.yml
ParidelPooya: Confirmed — `build.yml` triggers on pushes to both `main` and `v1.x`, and `GITHUB_BASE_REF` is empty on push events, so a push to `v1.x` diffed against `origin/main` and would attribute unrelated differences to that push.

Fixed in b09db048. I took the first of your two options: there is no meaningful `--changed` base for a push, so the step is now scoped to `github.event_name == 'pull_request'` and the `:-main` fallback is removed rather than left to guess. `origin/${GITHUB_REF_NAME}` would have resolved to the pushed branch itself, so the diff would have been empty — harmless but pointless.

biome.jsonc
ParidelPooya: Not intentional — a real gap, and you are right about the cause. The deleted rule checked the filename (index.js:123) *and* every directory segment (index.js:146-164). Biome's `useFilenamingConvention` only does the filename; verified that a kebab-case file inside `BadDirName/nested_dir/` produces no diagnostic while `BadFileName.ts` errors.

Fixed in b09db048. The repo already contained `scripts/check-filenames.js`, which checks directory segments — but it was wired to nothing, and it failed on the current tree because it flagged Jest's `__mocks__`, which is likely why it was never adopted. Added that exemption (the same one the deleted ESLint config and the biome.jsonc override carry), wired it up as `lint:filenames`, and added a step to the lint job. Verified it exits 1 on a non-kebab directory and 0 on the current tree.

`biome.jsonc` now records the split coverage next to the override, including that the two halves must agree on `__mocks__`.

yaythomas (approved):

PR #893, opened 2026-09-02
refactor(insight): memoise llmProviderOptions so its effect dependency is checkable

ParidelPooya:
`useExhaustiveDependencies` reported that the settings effect reads
`llmProviderOptions.some` and `llmProviderOptions[0].value` without listing them.
It was suppressed during the Biome migration with a note calling it a
stale-closure defect. That was overstated: the list is derived purely from
`capabilities.copilot` and `capabilities.localLlm`, both of which WERE in the
dependency array, so the effect already re-ran whenever the options changed.
There was no live bug.

What was real is that the dependency was indirect and therefore unverifiable. A
future input to `llmProviderOptions` -- another capability flag, say -- could be
added without a matching entry in the effect's dependencies, and the effect would
silently start acting on a stale list with nothing to catch it.

Memoising the list on the two flags it actually derives from fixes that. Identity
now changes exactly when those flags change, which is exactly when the effect used
to re-run, so behaviour is unchanged; the difference is that the dependency is
declared where a tool can check it. Adding a third input to the list without
declaring it now fails lint.

The suppression is not removed but its scope shrinks to the one finding that is
genuinely an idiom: `visible` is an intentional extra dependency, never read in
the body, whose only job is to re-run the effect when the modal reopens so
unsaved edits are discarded and the form resets from `settings`. Removing it
would silently start preserving stale edits across opens, and the rule has no way
to express "re-run on this signal". The reason text now says that instead of
claiming a defect.

Verified: no unsuppressed diagnostics remain on the file, both suppressions are
still live (removing them reproduces exactly the two `visible` findings and
nothing else), `tsc --noEmit` is clean, and the webview bundle builds.

### DIFF
--- a/packages/aws-durable-execution-sdk-js-insight-vscode/webview-ui/src/SettingsModal.tsx
+++ b/packages/aws-durable-execution-sdk-js-insight-vscode/webview-ui/src/SettingsModal.tsx
@@ -107,13 +107,18 @@ export function SettingsModal({
   // dropdown cannot show a label for a choice it doesn't offer. Copilot is
   // omitted entirely where the host can't reach it rather than shown disabled:
   // there is no action the user could take to enable it.
-  // Memoised so the effect below can depend on it directly. The list is derived
-  // purely from these two capability flags, so the identity changes exactly when
-  // they do -- which is exactly when the effect used to re-run via its old
-  // `capabilities.copilot, capabilities.localLlm` dependencies. Behaviour is
-  // unchanged; what changes is that the dependency is now machine-checkable, so a
-  // future input to this list cannot be added without either declaring it here or
-  // failing lint.
+  //
+  // Memoised so the effect below depends on this list by identity instead of on the
+  // flags behind it. That is what makes the dependency checkable: a third input
+  // cannot be added here without declaring it or failing lint.
+  //
+  // The identity is stable in practice, not by contract. React documents useMemo as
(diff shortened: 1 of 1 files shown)


ParidelPooya:
Both substantive points fixed in 7ed85b99.

**The stale sibling reason** — you're right that this mattered more than a normal comment nit. The reason was written against the *previous* state of the finding above it, and memoising is exactly what invalidated it. Both are now self-contained and symmetric: each says what `visible` is for and points at the other as the same idiom. Confirmed the two stripped findings are identical in kind (`specifies more dependencies than necessary: visible` ×2).

**The useMemo guarantee** — accepted, and the comment now carries the shape of it: identity is stable *in practice, not by contract*; a dropped cache would re-run the effect and reset the form, losing unsaved edits, which the previous version could not do since the rebuilt array wasn't a dependency; eviction is tied to Activity/Offscreen unmounting, which this tree doesn't use. Agreed the approach stands — memoising is the only route to the checkability.

Also took the trim: the block comment no longer restates the commit message.

**Deferred, as you suggested:** the non-empty tuple typing for `llmProviderOptions[0]` (pre-existing, own change) and a test for the `visible` invariant (no React component tests or `@testing-library` in the repo, so it needs new infrastructure).

On the clone note — thanks, and no action needed: I'm working in the original clone and haven't touched `node_modules`. This branch changes no dependencies, so your symlinks stay valid.

yaythomas (approved):

PR #894, opened 2026-09-02
fix(sdk): record per-item completion so summarized replay rebuilds FLAT map items

ParidelPooya:
`context.map` and `context.parallel` with `nesting: FLAT` rebuilt an **empty** result on replay when the batch was large enough to be summarized, even though every item had succeeded.

A batch whose aggregate result exceeds the checkpoint size limit is stored as a summary with `ReplayChildren`, and each later resumption reconstructs it from the per-item checkpoints in `ConcurrencyController.replayItems`. That reconstruction decided whether an item had finished by probing the item's context checkpoint — but with `FLAT` nesting the per-item contexts are virtual and never checkpointed, so every item read as unfinished and was skipped.

Not cosmetic: code deriving control flow from the batch result, such as a fan-out sized from the results, then creates a different set of operations on replay than it did live. The replay-consistency check detects the divergence and raises `NonDeterministicExecutionError`, which SDKs at or below 2.3.0 reported as a `PENDING` response rather than a failure. While other work is still draining `PENDING` is a legal answer, so the corruption is invisible; once nothing is left pending the service rejects the response with "Cannot return PENDING status with no pending operations" and the execution fails after deterministic retries. Batches whose result fits in one checkpoint are replayed by deserialization and never enter this path.

## Change

The batch payload now records which items reached a terminal state, one character per item in index order. Replay reads that record instead of inferring it. Only terminality is recorded — an item's `SUCCEEDED`/`FAILED` outcome still comes from re-driving it.

Terminality resolves in order of authority:

1. the recorded statuses, with any context record the item has required to agree
2. the item's own context checkpoint, for payloads with no recorded statuses
3. a probe of the item's operations, for payloads predating the field

The agreement requirement in (1) is one-sided on purpose. A marker saying unfinished overrides a terminal context record, because the statuses are what the live `BatchResult` contained. The reverse is not honoured: re-driving an item whose context checkpoint is non-terminal hits `checkForNonResolvingPromise` and the reconstruction never settles (#751).

The probe in (3) walks every recorded operation and requires all of them to be terminal. Reading only the first is unsound — a mid-flight item can have a `SUCCEEDED` first operation, and re-driving it would execute its unfinished operations for real.

Two payload shapes are recognised, since `ReplayChildren` is reached two ways: the summary envelope carrying `itemStatuses`, and the full serialized result written when `childOperationsDepth` opts a context into child preservation, which has no summary envelope but does carry `all: [{ index, status }]`. Parsing is validated as strictly as `parseRecordedCompletionReason`.

There is deliberately no size handling on the record. Every batch item runs its body inside a context, and a context is meant to wrap at least one durable operation, so each item costs at least one. The service caps an execution at 30,000 operations, which bounds any batch at roughly that many items — about 30 KB of markers at one byte each, an order of magnitude inside the 256 KB checkpoint limit. Capping the marker count instead would be actively harmful: dropping the field sends replay back to probing, and probing cannot recover an item whose body created no durable operation, which is the corruption this record exists to prevent. A test pins the bound at the largest batch the service allows.

## Behaviour change

A virtual item with no durable operation of its own is now re-driven on replay, so any part of its mapper body not wrapped in a durable operation runs again. This is the documented contract for virtual contexts and what `NESTED` already does, but it is new on this path.

That shape is against best practice regardless — a context is a container for durable operations, and one with nothing durable inside it is not checkpointed, so its work re-runs on every replay anyway. The TSDoc for `DurableContext.runInChildContext` and `MapFunc` now says so, and #896 tracks an eslint rule for it.

## Testing

New `examples/map/flat-summarized-replay` covering the rebuild across a suspension for `FLAT` and `NESTED`, an item that checkpoints no durable operation, and a small-result control. It is `localOnly`: the suspension depends on the local runner's virtual clock, so deploying it would only produce a skipped cloud test. Every case asserts at least two invocations, so none can pass without actually suspending.

Against unpatched code both `FLAT` cases fail (`replayedResultCount` 0 against a live count of 8) while the controls pass. 14 new unit cases cover both directions of a statuses/record disagreement, the `childOperationsDepth` shape, a crafted `itemStatuses` value, the probe rejecting a mid-flight item, and the size bound at the service's operation limit.

The `PENDING` misreport that hid this was fixed separately on mainline by #867; this addresses the underlying corruption.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- A `context.map` or `context.parallel` using `nesting: FLAT` whose aggregate result exceeded the
+  256KB checkpoint limit rebuilt an **empty** result on replay, even though every item had
+  succeeded. Such a batch is checkpointed as a summary and reconstructed from its per-item
+  checkpoints on each later resumption, but the reconstruction decided whether an item had finished
+  by probing the item's context checkpoint — and with `FLAT` nesting those per-item contexts are
+  virtual and never checkpointed, so every item read as unfinished and was skipped.
+
+  Code deriving control flow from the batch result (for example a fan-out sized from the results)
+  therefore created a different set of operations on replay than it had live. The replay-consistency
+  check detected the divergence and raised `NonDeterministicExecutionError`, which SDKs at or below
+  2.3.0 reported as a `PENDING` response rather than a failure; once nothing was left pending the
+  service rejected the response with "Cannot return PENDING status with no pending operations" and,
+  after deterministic retries, failed the execution.
+
+  The batch payload now records which items reached a terminal state, and replay reads that record
+  instead of inferring it. Only large batches were affected: a result that fits in a single
+  checkpoint is replayed by deserialization and never enters the reconstruction path.
+
+  One behaviour change on this path: a virtual item that performs no durable operation of its own is
+  now re-driven during replay, so any part of its mapper body not wrapped in a durable operation
+  runs again. This matches the documented contract for virtual contexts and what `NESTED` already
+  does. Such a body is against best practice in any case — a context is a container for durable
+  operations — and the TSDoc for `runInChildContext` and `MapFunc` now says so.
+
 ### Added
 
 - Support for JavaScript runtimes that do not provide `async_hooks.AsyncLocalStorage` or
--- a/packages/aws-durable-execution-sdk-js-examples/src/examples/map/flat-summarized-replay/map-flat-summarized-replay.test.ts
+++ b/packages/aws-durable-execution-sdk-js-examples/src/examples/map/flat-summarized-replay/map-flat-summarized-replay.test.ts
@@ -1,0 +1,129 @@
+import { LocalDurableTestRunner } from "@aws/durable-execution-sdk-js-testing";
+import { NestingType } from "@aws/durable-execution-sdk-js";
+import { handler } from "./map-flat-summarized-replay";
+
+const ITEM_COUNT = 8;
+
+/**
+ * A map whose aggregate result exceeds the 256KB checkpoint limit is
+ * checkpointed as a summary with ContextOptions.ReplayChildren, and on every
+ * later resumption is rebuilt from its child checkpoints by
+ * ConcurrencyController.replayItems.
+ *
+ * replayItems decided whether each item finished by probing the item's context
+ * checkpoint (`${mapId}-${n}`). With `nesting: FLAT` those per-item contexts are
+ * VIRTUAL and never checkpointed, so the probe found nothing, every item was
+ * treated as non-terminal and skipped, and the rebuilt BatchResult came back
+ * EMPTY even though all items had succeeded.
+ *
+ * Why that is severe rather than cosmetic: code deriving control flow from the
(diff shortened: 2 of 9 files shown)


yaythomas (commented):
summary-generators.ts:20,64 allows 262,144 status markers, but the complete summary JSON is 262,297 bytes, which 153 bytes over the documented 256 KB maximum. The cap should reserve envelope headroom or check the final serialized size. 

This only affects roughly quarter-million-item batches.

ParidelPooya:
Good catch, and the arithmetic is exact — I reproduced 262,297 bytes against the 262,144 limit.

Fixed by checking the finished payload rather than reserving headroom, because a marker cap can only ever approximate the envelope, and `composeSummaryGenerator` appends a custom `summaryGenerator`'s output afterwards, which can be any length. `stringifyWithinLimit` serializes the record and, if it exceeds `CHECKPOINT_SIZE_LIMIT_BYTES`, drops `itemStatuses` and re-serializes. Replay then falls back to probing, which is the degraded-but-correct path.

Where a custom summary is what pushes it over, the customer's string is kept and `itemStatuses` is dropped — the custom value is the documented contract, ours is an optimisation.

The old cap survives only as a cheap allocation guard (one byte per marker, so a run longer than the whole limit cannot fit inside any envelope) and is now expressed in terms of `CHECKPOINT_SIZE_LIMIT_BYTES` rather than a literal.

Three tests, asserting `Buffer.byteLength` of the finished payload:
- markers exactly at the limit -> payload within limit, `itemStatuses` absent, `completionReason` retained
- 1,000 items -> markers present
- large custom summary -> markers dropped, custom string intact

The first one fails at 262,297 against the previous behaviour, so it pins your case directly.

yaythomas:
Dropping itemStatuses reintroduces the original replay corruption

summary-generators.ts:22-36 says dropping the field is safe because replay probes checkpoints. However, concurrent-execution-handler.ts:445-446 explicitly documents that probing cannot recover a completed FLAT item whose mapper created no durable operations.

I reproduced this with a valid dense 262,144-item FLAT batch:

Live result: 262,144 successful items
Final summary drops itemStatuses to remain under 256 KB
Replay finds no item contexts or child operations
Rebuilt result: 0 items

this replacement creates the same correctness failure at the size boundary.

ParidelPooya:
Removed the size handling rather than fixing it — you were right that dropping `itemStatuses` just trades one corruption for another, since probing cannot recover an item whose body created no durable operation.

The reason no size handling is needed: every batch item runs its body inside a context, and a context is a container for durable operations, so each item costs at least one. The service caps an execution at 30,000 operations, which bounds any batch at roughly that many items — about 30 KB of markers at one byte each, an order of magnitude inside the 256 KB checkpoint limit. A test pins that at the limit rather than leaving it as a comment.

For completeness on the shape that motivated the cap: a mapper doing nothing durable does cost 0 operations per item (both the START and SUCCEED checkpoints are skipped for a virtual context, so a pure-transform FLAT map is ~1 operation regardless of item count, and I measured 4 operations for 8 items versus 12 with a step). So the operation limit does not bound item count there. But that shape is against best practice — the work is not checkpointed and re-runs on every replay anyway — so it should not shape the design. The TSDoc for `runInChildContext` and `MapFunc` now documents the rule, and #896 tracks an eslint rule to enforce it.

For the record, the exact boundary was 261,992 items, not 262,144: the markers share the payload with a 153-byte envelope, which is why the old cap sat 152 items above the real threshold.

yaythomas (approved):

PR #914, opened 2026-09-11
fix(otel): emit type declarations on every build

nvasiu:
*Issue #, if available:*

N/A

*Description of changes:*

### Problem
`@aws/durable-execution-sdk-js-otel@1.0.0` shipped with the `/dist-types/` folder missing.

### Root Cause
- `tsconfig.build.json` set `incremental: true`.
- The package prebuild runs `rm -rf dist dist-cjs dist-types` but does not remove `tsbuildinfo`.
  - `tsbuildinfo` is a cache that TS writes when compiling with `incremental: true`.
- On the publish rebuild the stale cache was used, no changes were detected and the build emit was skipped. So the missing `/dist-types/` was not rebuilt.

### Fix
Removing `incremental: true` from `tsconfig.build.json` forces the types to always build.

Verified by reproducing the build and rebuild. When `incremental: true` was present, the second build emitted no type files. But with `incremental` removed, all type files are emitted every time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,6 +67,8 @@ jobs:
         run: npm ci
       - name: Build project
         run: npm run build
+      - name: Verify every package ships the files its package.json declares
+        run: node .github/workflows/scripts/test-publish/verify-declared-files-shipped.mjs
       - name: Upload build artifacts
         uses: actions/upload-artifact@v7
         with:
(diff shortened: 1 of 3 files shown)


yaythomas:
Could we add a regression test that reproduces the release sequence? It should build the OTEL package twice, run `npm pack --dry-run`, and verify that every `types` entry from `package.json` exists in the packed files. The old configuration would fail because the second build omits `dist-types`; this change should pass.

Then run it in the build workflow after npm ci.

nvasiu:
Added a regression test that verifies a build includes all the package's declared files. This test runs on all packages, not just Otel.

Ran the test on the Otel package without this PR's fix, and it caught the missing `/dist-types/`. After applying the fix, the test passes successfully.

yaythomas (approved):
:shipit:

PR #921, opened 2026-09-15
fix(insight): queue exports per execution so concurrent executions never drop records

### LINKED ISSUE #916: [Feature]: Workflow Insight scheduler drops records across concurrent invocations
### What would you like?

Description

ExportScheduler uses one global pending slot. Records from different execution ARNs can overwrite each other, even though they are unrelated snapshots.

### Possible Implementation

- Records from different execution ARNs must never replace each other.
- Terminal records must always reach the exporter before drain completes.
- Burst coalescing must be scoped and documented.

### Is this a breaking change?

No

### Does this require an RFC?

No

### Additional Context

_No response_

wangyb-A:
## Problem

`ExportScheduler` in the Workflow Insight plugin has one pending slot shared by every execution the plugin instance serves. When one instance handles more than one execution at a time (the local test runtime, or any host that runs several invocations in one process), a record scheduled by execution A while an export is in flight replaces a pending record from execution B. B's `wrapInvocation` then awaits the shared in-flight promise and returns as if its record had been delivered. Terminal records are lost silently, in every `emitMode`.

## What changed

Pending records are keyed by `executionArn` in an insertion-ordered `Map`.

- Snapshots of the **same** execution still coalesce: the latest one wins.
- Records of **different** executions never replace each other. They are exported oldest-first, and a re-scheduled execution keeps its place in line, so a chatty execution cannot starve the others.
- `drain(executionArn)` waits only for that execution's pending or in-flight record. `wrapInvocation` passes its own ARN.
- Exports stay strictly serial. Exporters still never see overlapping `export()` calls.

`ExportScheduler` stays in `src/index.ts`; no public API or exporter contract changes. The README's "How It Works" paragraph now states that coalescing is per execution.

## Tests

Six tests in `src/index.test.ts`, driven through the plugin hooks with a gated exporter whose `export()` can be held open:

- cross-execution overwrite (the #916 scenario)
- same-execution coalescing is preserved
- per-execution coalescing keeps first-scheduled order
- 50 concurrent invocations each deliver their terminal record, with max one `export()` in flight
- `drain` returns for the invoking execution without waiting on another execution's export
- a failing export for one execution does not block another

Four of these fail against the previous scheduler (verified by running them with the old `index.ts` in place); the other two guard behaviour that must not regress.

`tsc --noEmit`, `jest` (87 tests in the package), `biome check`, and `npm run build` for the package are green locally.

## Manual verification

N/A: the concurrent-invocation case is reproduced deterministically in unit tests.

Fixes #916


ParidelPooya (changes requested):
A synchronous throw from exporter code wedges the scheduler and hangs the next invocation

   The per-execution queue itself looks correct, and I confirmed it fixes #916. This is about one line the PR rewrites.

   `render` and `export` are user code. The pump calls them inside
   [`this.exporters.map((exporter) => exporter.export(truncateRecord(record, …,
   exporter.render…)))`](https://github.com/aws/aws-durable-execution-sdk-js/blob/f5125f8041214c14c106133eef44025681bf99c2/packages/aws-durable-execution-sdk-js-insight/src/index.ts#
   L446).
   The callback is not `async`, and `truncateRecord` calls `render` synchronously. So a synchronous throw propagates out of
   `.map()` instead of becoming a rejected promise, and `Promise.allSettled` never sees it.

   The chain from there:

   1. The throw happens during the pump's synchronous prologue, before its first `await`.
   2. So the pump's `finally` runs synchronously and sets `this.inFlight = undefined`.
   3. `schedule` then assigns `this.inFlight = <the rejected promise>`, because that assignment happens after `pump()` returns.
   4. So `inFlight` holds a settled rejected promise that nothing ever clears.
   5. `schedule` starts a pump only when `inFlight === undefined`. So no later record is ever exported; they accumulate in `pending`.
   6. `drain(arn)` sees the ARN pending, registers a waiter, and no pump exists to settle it. So the promise never resolves and `wrapInvocation` never returns.

   Measured against the built `dist/index.mjs`, with an exporter whose `render` throws on the first record only, then a
   second execution scheduling a healthy record:

   ```text
   main (b16eafb2)   a_drain="rejected: Error: render blew up"  b_drain="rejected: Error: render blew up"  b_drain_ms=0     b_record_exported=false
   this PR (f5125f8) a_drain="resolved"                         b_drain="hung >2000ms"                     b_drain_ms=2011  b_record_exported=false

   main is broken too, so this is not a regression introduced here. The failure mode changes, and the new one is worse on Lambda: main fails the next invocation with the exporter's
   error, while this revision hangs it until the function times out, having exported nothing and reported nothing. With Node's default unhandled-rejection setting and no host
   listener, both branches exit the process with status 1 on the first throw; the divergence above is what a host that installs an unhandledRejection listener sees, which the Lambda
   Node runtime does. Either way, neither branch honors the documented contract that an exporter error never fails the execution.

   Making the map callback async turns the synchronous throw into a rejected promise, which allSettled swallows:

   this.exporters.map(async (exporter) =>
     exporter.export(
       truncateRecord(record, exporter.maxRecordSizeBytes, exporter.render?.bind(exporter)),
     ),
   )

   I applied exactly that change, rebuilt, and re-ran the probe:

   with fix          a_drain="resolved"  b_drain="resolved"  b_drain_ms=0  b_record_exported=true  unhandled_rejections=0

   Please add a regression test with it. The existing "a failing export for one execution does not block the others" test throws from an async method, so it covers only the
   rejected-promise path. A test needs either a render that throws or a non-async export that throws, and it should assert that a later execution's record is still exported and that
   its wrapInvocation returns. ```

